PR #23, opened 2024-06-11
Rework Core: scheduler and state machine

Ibrassow:
Main changes on the scheduler (previously tasko), state machine.
Introduced Core module for the framework.
A little bit of cleaning. 
Need more features in the future but ok for now.

Will pull from the linting branch after its merge


flight/core/scheduler/loop.py
```diff
@@ -165,16 +166,22 @@ def __init__(self, debug=False):
...
+
     def add_task(self, awaitable_task, priority):
         """
         Add a concurrent task (known as a coroutine, implemented as a generator in CircuitPython)
         Use:
           scheduler.add_task( my_async_method() )
         :param awaitable_task:  The coroutine to be concurrently driven to completion.
         """
-        self._debug("adding task ", awaitable_task)
+        if self.debug:
```
tomyyyD, line 181: It seems unnecessary to me to have 
`if self.debug: self._debug("...")`
as `self._debug` will only print if self.debug is True as it is set to the "do nothing" lambda function on line 167 if self.debug is false.

tomyyyD: basically `self._debug` will only ever print if `self.debug` is true so there's no point in checking if `self.debug` is True.

tests/test_scheduler.py
```diff
@@ -0,0 +1 @@
+# TODO
```
tomyyyD, line 1: Why wait to do these? Do you want the emulator first?

tests/test_state_manager.py
```diff
@@ -0,0 +1 @@
+# TODO
```
tomyyyD, line 1: same as comment on test_scheduler.py

flight/apps/jetson_comms/message_id.py
```diff
@@ -59,7 +59,9 @@ class Identifiers:
     REQUEST_AOD_LAST_ESTIMATE = 0x51
     REQUEST_ATTITUDE_ESTIMATE = 0x52
     REQUEST_ORBIT_ESTIMATE = 0x53
-    RESOLVE_STATE = 0x54  # Payload simulate forward to current time (prediction)
+    RESOLVE_STATE = (
+        0x54  # Payload simulate forward to current time (prediction)
```
tomyyyD, line 63: Why wrap this single value in parentheses?

tomyyyD (approved):
For the most part, looks really good. Some stuff in the tasko loop that you should look at the with debug logging, and preferably we write the two empty unit test files before we merge, but if you're waiting on the emulator for that, then its reasonable to keep them blank for now.

flight/apps/jetson_comms/message_id.py
Ibrassow: :) they were part of the previous code from the semester, not this PR. One of the pre-commit tools probably changed it. Funny.

flight/core/scheduler/loop.py
Ibrassow: Did a benchmark on the board and checking a flag is faster than the no-op lambda (surprisingly). This was the reason for the addition of the flag, esp for flight code where the scheduler must be as fast as possible. 

Now, the self.debug attribute fct is useless and can be replaced by print statements only, which will be done in the next commit.

PR #24, opened 2024-06-12
Fixed SunSensor unsupported types error

sophizhao:
Updated sun_sensor driver and the emulator.py code to save lux readings as numerical values.

### DIFF
--- a/emulator/drivers/sun_sensor.py
+++ b/emulator/drivers/sun_sensor.py
@@ -1,7 +1,17 @@
 class SunSensor:
-    def __init__(self, lux) -> None:
-        self.__lux = lux
+    def __init__(self, x, y, z) -> None:
+        self.__x = x
+        self.__y = y
+        self.__z = z
 
     @property
-    def lux(self):
-        return self.__lux
+    def x(self):
+        return self.__x
+
+    @property
+    def y(self):
+        return self.__y
+
+    @property
+    def z(self):
+        return self.__z
--- a/emulator/drivers/sun_sensor.py
+++ b/emulator/drivers/sun_sensor.py
@@ -1,17 +1,7 @@
 class SunSensor:
-    def __init__(self, x, y, z) -> None:
-        self.__x = x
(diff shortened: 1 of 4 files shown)


flight/apps/data_handler.py
```diff
@@ -93,7 +93,7 @@ def __init__(
         persistent: bool = True,
         line_limit: int = 1000,
         new_config_file: bool = False,
-        home_path: str = "/sd",
```
tomyyyD, line 96: These all need to get changed back to /sd as we want to make sure hardware is working. Will come up with a more permanent fix for this later.

tomyyyD: same with al other changed instances in this file

emulator/emulator.py
```diff
@@ -51,12 +51,15 @@ def __init__(self, enable_middleware, debug) -> None:
         self._gps = None
         self._charger = None
 
-        self._sun_sensor_xp = SunSensor(900)
-        self._sun_sensor_xm = SunSensor(48000)
-        self._sun_sensor_yp = SunSensor(85000)
-        self._sun_sensor_ym = SunSensor(200)
-        self._sun_sensor_zp = SunSensor(12000)
-        self._sun_sensor_zm = SunSensor(5000)
+        sun_sensor_p = SunSensor(900,85000,12000)
```
tomyyyD, line 54: I would prefer to keep each sun sensor as its own device to make sure we are staying as true as we can be to the actual hardware experience. This won't affect us now, but it may affect things in the future

emulator/drivers/sun_sensor.py
```diff
@@ -1,7 +1,17 @@
 class SunSensor:
-    def __init__(self, lux) -> None:
-        self.__lux = lux
+    def __init__(self, x, y, z) -> None:
```
tomyyyD, line 2: I would like each sensor to be independent of the others with a lux value that returns if you could find a way to do this

PR #28, opened 2024-06-13
Emulator

tomyyyD:
add emulator, and some more middleware updates.


emulator/drivers/middleware/generic_driver.py
```diff
@@ -0,0 +1,113 @@
...
+        isProp, checker, m_exception = self.handleable[method.__name__]
+
+        def handle(*args, **kwargs):
+            try:
+                res = method(*args, **kwargs)
+                flags = self.get_flags
+                if checker(res, flags):
+                    return res
+                else:
+                    raise m_exception("erroneus result")
```
Ibrassow, line 40: would show the erroneous value in the exception raise message at this stage

emulator/drivers/sun_sensor.py
```diff
@@ -0,0 +1,7 @@
+class SunSensor:
```
Ibrassow, line 1: Need to rename everywhere (in the whole codebase) SunSensor by LightSensor (the sun sensor is the whole, or at least 3 OPT4001)

emulator/configuration.py
```diff
@@ -0,0 +1,9 @@
+from .cubesat import CubeSat
```
Ibrassow, line 1: careful with relative imports (it can quickly become messy). would prefer absolute import for everything

emulator/cubesat.py
```diff
@@ -0,0 +1,227 @@
+from .drivers.diagnostics.diagnostics import Diagnostics
```
Ibrassow, line 1: same comment for relative imports (you can extend it to the whole)

PR #34, opened 2024-06-30
Sun Vector Module

Ibrassow:


flight/hal/drivers/opt4001.py
```diff
@@ -213,7 +207,7 @@ def __init__(
         package: int = 0,
         quick_wakeup: bool = False,
         lux_range: int = 0b1100,
-        conversion_time: int = 0b1000,
+        conversion_time: int = 0b0000,
```
tomyyyD, line 210: no reason to change the driver default values. Instead just input the desired value on instantiation of our devices. Default value should not be the the lowest possible conversion time.

flight/tasks/sun.py
```diff
@@ -2,21 +2,34 @@
...
+        "yp",
+        "ym",
+        "zp",
+    ]
 
-    # Fake starting sun vector
+    THRESHOLD_ILLUMINATION_LUX = 3000
+
+    status = False
     sun_vector = [1, 0, 0]
```
tomyyyD, line 33: why start with non-zero vector here?

tests/cp_mock.py
```diff
@@ -1,6 +1,13 @@
+import importlib
 import sys
 
+# Add necessary paths
 sys.path.append("emulator/cp/")
 sys.path.append("flight/")
+
+# Mock other modules
 sys.modules["micropython"] = __import__("micropython_mock")
```
tomyyyD, line 9: would rather you put these two sys.modules calls in emulator/cp_mock.py

tomyyyD (approved):
A few comments, I'm not entirely convinced that this solution will give us a vector even experiencing failures, but at least it won't break other tasks if a failure occurs.

tests/cp_mock.py
Ibrassow: Will keep the 2 cp_mock separate for now. We need to fix some emulator interfacing/path issues with the unit test pipeline first

flight/tasks/sun.py
Ibrassow: Good catch. Missed legacy code. Changed the initial status as well.

emulator/cubesat.py
```diff
@@ -138,46 +137,39 @@ def TORQUE_Z(self):
...
-    def SUN_SENSOR_ZP(self):
-        """SUN_SENSOR_ZP: Returns the sun sensor in the z+ direction
+    def LIGHT_SENSOR_ZP(self):
+        """LIGHT_SENSOR_ZP: Returns the light sensor in the z+ direction
         :return: object or None
         """
-        return self._sun_sensor_zp
-
-    @property
-    def SUN_SENSOR_ZM(self):
```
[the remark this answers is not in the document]
Ibrassow: Yes. Part of a larger configuration change for the sun acquisition. See Slack.

flight/apps/sun.py
```diff
@@ -1,34 +1,158 @@
...
+    # norm = MAX_RANGE
+
+    if norm == 0:  # Avoid division by zero - not perfect
+        status = SUN_VECTOR_STATUS.UNDETERMINED_VECTOR
+        return status, sun_body
 
-    # Normalize the vector if the norm is not zero
-    if norm != 0:
-        sun_body = [x / norm for x in sun_body]
+    sun_body[0] = sun_body[0] / norm
```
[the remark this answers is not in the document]
Ibrassow: good

flight/hal/drivers/opt4001.py
Ibrassow: Rolling back for now. Waiting time should be adapted based on conversion time

flight/apps/sun.py
```diff
@@ -1,34 +1,158 @@
...
     Returns:
+        eclipse (bool): True if the satellite is in eclipse, False otherwise
 
     """
-    # TODO
-    return False
+    eclipse = None
+
+    if raw_readings.count(ERROR_LUX) == NUM_LIGHT_SENSORS:
+        return eclipse
```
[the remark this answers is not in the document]
Ibrassow: Changed docs

PR #36, opened 2024-07-09
Fix HW freeze - Revert "Revert "Sun Vector Module""

Ibrassow:
Reverts cmu-spacecraft-design-build-fly-2023/FSW-mainboard#35


Ibrassow:
@tomyyyD: "During the software boot-up process, everything works as expected and we can copy over files just fine. Once the software reaches the nominal state, we can no longer make any changes to the code without the satellite freezing. Freezing occurs when any files are changed as well as a when new file is copied over, or a file is deleted. Freezing also occurs when you send a sigint signal (ctrl + c) to stop the flight software"

PR #46, opened 2024-09-04
Comms Protocol Refactoring

Akshat412:
Updated comms application and task for SAT to send TM frames as heartbeats and receive/decode GS ACKs.


Ibrassow (approved):
LGTM!

PR #47, opened 2024-09-26
Memory opti

Ibrassow:
The first series of changes to optimize the FSW

Changelogs 
- Data_limit in bytes rather than line limits
- Added write intervals, instead of writing every logs 
- Added frequency access to tasks
- Moved SM config and task registry locally
- Moved driver imports locally in interfaces


Akshat412 (approved):
Looks good

PR #52, opened 2024-11-04
Functional file TX to GS

Akshat412:
Changes:
- Packetized file TX
- Centralized comms state machine transitions


flight/tasks/comms.py
```diff
@@ -1,56 +1,112 @@
...
             self.TX_COUNT_THRESHOLD = int(self.frequency / self.TX_heartbeat_frequency)
+            self.TX_COUNTER = self.TX_COUNT_THRESHOLD - 1
             self.frequency_set = True
 
+            self.log_info(f"Heartbeat frequency threshold set to {self.TX_COUNT_THRESHOLD}")
+
         if SM.current_state == STATES.NOMINAL:
+            if not DH.data_process_exists("img"):
+                # TODO: Move image process to another task
+                DH.register_data_process("img", "b", True)
```
Ibrassow, line 53: agree with the TODO. I assume this was for testing purposes?

Ibrassow (changes requested):
I added initial unit tests for some of the comms code. Given the HW dependency, there is a lot of mocking. You should go over them and add/remove what makes sense.

PR #54, opened 2024-11-05
SIL

Ibrassow:
Changelogs

-   SIL interface running and working. For now, it's using Argusloop as a backend
-   Simulated devices: IMU, GPS, Light Sensors, Torque coil drivers
-   HAL interface modification for arrays of devices (light sensors and torque coils)
-   HAL interface for device availability

Notes for future

-   Simulation acceleration by scaling the FSW frequencies (right now, everything is "real-time")
-   Configuration test cases
-   Moving backend to C++ sim (@kmkar31)


kmkar31 (commented):
Looked through files. No major changes except:

1. Need units for sensor reading (Esp. weird ones like cm for GPS)
2. Need docstrings for functions in simulator.py

perrintong325 (approved):
for the interfaces we might have to do individual error codes instead of just 1 error code for the same components

Ibrassow:
> Looked through files. No major changes except:
> 
>     1. Need units for sensor reading (Esp. weird ones like cm for GPS)
> 
>     2. Need docstrings for functions in simulator.py

2. done

1. Will integrate all that once the proper GPS driver is finished.

PR #56, opened 2024-11-09
File TX, State Machine, and Error Handling

Akshat412:
Changes:
- Functional packetized file TX to the GS
- Comms SM with centralized state transitions
- Error handling for ground pass timeouts
- Error handling for missing CDH dependencies (filepaths etc.)
- Unit testing of state machine


flight/apps/comms/comms.py
```diff
@@ -37,20 +37,23 @@ class MSG_ID:
 
 
 class SATELLITE_RADIO:
+    # Hardware abstraction for satellite
+    sat = SATELLITE
```
Ibrassow, line 41: would remove that reference and use directly SATELLITE in this file. This would be clear that this comes from the HAL

flight/apps/comms/comms.py
```diff
@@ -293,17 +353,16 @@ def transmit_message(cls):
...
         else:
-            # Transmit SAT heartbeat
-            cls.state = COMMS_STATE.TX_HEARTBEAT
+            # Unknown state, just send
+            logger.warning(f"[COMMS ERROR] SAT received {cls.gs_rq_message_ID}")
             cls.tx_message = cls.tm_frame
 
         # Send a message to GS
-        SATELLITE.RADIO.send(cls.tx_message)
+        cls.sat.RADIO.send(cls.tx_message)
```
Ibrassow, line 364: same note

Ibrassow (approved):

PR #57, opened 2024-11-14
GPS Driver Update

WalkerAdi:
Changed the driver from the adafruit one to one that I wrote. The basic difference is that the GPS module is changed from NMEA messages to the SkyTraq Binary message protocol, which has the navigation data message Argus-1 was built off of.


Akshat412:
Tabling this to Spring 25

PR #59, opened 2024-11-17
Mcm Allocation and Controllers

Herpderk:
Changelogs 

ADCS
- Magnetic torque coil voltage allocator 
- Controllers implementation
- Mode management implementation 

CD&H
- Availability and clearing capabilities to DH
- Scheduler trim and documentation

TODO:
1. hardware sanity check


flight/apps/adcs/mcm.py
```diff
@@ -7,4 +7,236 @@
...
+        0.0, -v[2], v[1],
+        v[2], 0.0, -v[0],
+        -v[1], v[0], 0.0,
+    ])
+
+
+'''
+Template controllers for spin stabilizing and sun pointing.
+'''
+class SpinStabilizingController(ABC):
```
Ibrassow, line 32: same comment for abstract stuff

Ibrassow (changes requested):
overall, I'm curious about the total memory footprint if yo could measure that but looks good overall

flight/apps/adcs/mcm.py
```diff
@@ -1,10 +1,140 @@
...
+    ang_vel_target = np.linalg.norm(ang_vel_reference)
+
+    @classmethod
+    def update_max_dipole_moment(self) -> None:
+        u_max = np.zeros(3)
+        for row in MagneticCoilAllocator.mat:
+            if not np.all(row == 0.0):
+                row_norm = np.linalg.norm(row)
+                u_max += row / row_norm * (MagnetorquerConst.V_MAX / MagnetorquerConst.V_CONVERT)
+        self.u_max = u_max
```
Herpderk, line 72: I think there's a more efficient way of doing this part than looping through the matrix, but I haven't quite figured out what it is.

Ibrassow: one thing already is to precompute (MagnetorquerConst.V_MAX / MagnetorquerConst.V_CONVERT) into a multiplicative constant (division are extremely expensive)

flight/apps/adcs/mcm.py
```diff
@@ -1,10 +1,140 @@
...
+    if spin_axis[np.argmax(np.abs(spin_axis))] < 0:
+        spin_axis = -spin_axis
+
+    # References and targets
+    ang_vel_reference = spin_axis * MCMConst.REF_FACTOR * ModeConst.STABLE_TOL
+    ang_vel_target = np.linalg.norm(ang_vel_reference)
+
+    @classmethod
+    def update_max_dipole_moment(self) -> None:
+        u_max = np.zeros(3)
```
Ibrassow, line 70: u_max[:] = 0  # Reset values to zero in place

flight/apps/adcs/mcm.py
Ibrassow: precompute (MagnetorquerConst.V_MAX / MagnetorquerConst.V_CONVERT) into a multiplicative constant, divisions are expensive

Ibrassow: might be worthwile for us to measure the processing speed of this (using the nanosec clock)

tests/test_mcm.py
```diff
@@ -0,0 +1,65 @@
...
+    )
+    assert dipole_moment.shape == (3,)
+
+
+def test_sun_pointing(readings) -> None:
+    dipole_moment = get_sun_pointing_dipole_moment(
+        readings["sun_vector"],
+        readings["pointing_error"],
+    )
+    assert dipole_moment.shape == (3,)
```
Ibrassow, line 42: Here, test the shape output is good but you want to test the computation as well with a clear input value (instead of random) and compare it to your reference output to make sure the computation is correct. 

You want also to test nominal cases and corner cases (near zero) or when a zero norm appears somewhere.

Be paranoid

Ibrassow (changes requested):

flight/apps/adcs/mcm.py
Herpderk: this is initializing a new u_max before assigning it to the class attribute, so there's nothing to reset. Unless you want me to directly update the class attribute?

Ibrassow (approved):
raising an issue for the tests to be fixed in another PR

PR #68, opened 2025-01-14
CDH: (Major) Core improvements, redesign & simplifications

Ibrassow:
Changelogs: this PR implements a range of changes following a SW ConOps redesign, necessary features for subsequent development (e.g. watchdogs, SM), and some fixes and optimizations.

State Machine
- Implemented new state machine following SW ConOps redesign (STARTUP, DETUMBLING, NOMINAL, LOW-POWER, PAYLOAD)
- Added time duration since the latest state change 
- Implemented DETUMBLING exit state transition, from both ADCS check (through DH) and timeout
- Set Detumbling Error Flag with timeout

Onboard Data Handling
- Ability to retrieve buffer data from the latest file (by default) to restore the full spacecraft state through all data processes (Retrieval with proper exception handling in case of failure)
- Now able to append to the latest file instead of creating a new one in the data process after every reboot
- Error handling for SD card path failure and check interface for the communication subsystem

Scheduler
- Constant task set across states (fixed frequencies and priorities), removing rescheduling during state transitions
- Simplified task configuration

Commands
- Removed enable/disable devices and tasks commands, as not needed (design decision)
- Updated the whole set of commands
- Some TM frame modification 

Build Pipeline
- Fix broken emulator from the HAL v2 PR (wasn't adapted)
- Removed redundant call to standard build for nothing) 
- Removed deletion of sd folder in emulation/simulation to properly simulate a reboot (and thus state retrieval)

ADCS
- Removal of star tracker since it doesn't (and won't exist)
- Slight cleaning


Ibrassow:
Still needs a HW test and some modification on the command list and processing.

Ibrassow:
Implementing the commands and the other TM frames will be handled in a subsequent PR.

flight/core/state_machine.py
```diff
@@ -38,33 +43,35 @@ def current_state(self):
...
         """
-        from core.sm_configuration import SM_CONFIGURATION
 
-        self.__config = SM_CONFIGURATION
+        from core.task_configuration import TASK_CONFIG
 
-        # TODO Validate the configuration and registry
-        self.__states = list(self.__config.keys())
+        self.__task_config = TASK_CONFIG
+        self.__states = [STATES.STARTUP, STATES.DETUMBLING, STATES.NOMINAL, STATES.LOW_POWER]
```
s026134, line 61: should the STATES.PAYLOAD be included in this as well or will this be added later?

emulator/emulator.py
```diff
@@ -68,6 +68,8 @@ def __init__(self, enable_middleware: bool, debug: bool, simulator, use_socket)
 
         self._jetson_power_monitor = self.init_device(PowerMonitor(4, 0.05))
         self._board_power_monitor = self.init_device(PowerMonitor(7.6, 0.1))
+        self._power_monitors["BOARD"] = self._jetson_power_monitor
```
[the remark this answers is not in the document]
Ibrassow: good catch

flight/core/state_machine.py
Ibrassow: done

flight/apps/comms/comms.py
```diff
@@ -232,7 +232,7 @@ def file_pack_metadata(cls):
 
     @classmethod
     def data_available(cls):
-        return SATELLITE.RADIO.rx_available()
+        return SATELLITE.RADIO.RX_available()
```
Akshat412, line 235: Also needs a change in the SX1262 driver, I'll do that on my branch or you can do it in this PR.

flight/core/states.py
```diff
@@ -25,21 +27,33 @@
...
-    LOW_POWER = const(0x03)
-    SAFE = const(0x04)
+    DETUMBLING = const(0x01)
+    NOMINAL = const(0x02)
+    PAYLOAD = const(0x03)
+    LOW_POWER = const(0x04)
+
+    TRANSITIONS = {
+        STARTUP: [DETUMBLING],
+        DETUMBLING: [NOMINAL, LOW_POWER],
```
Akshat412, line 50: Do we need a transition directly from DETUMBLING to LOW_POWER? Also, LOW_POWER doesn't currently have a transition back to DETUMBLING.

I made some changes to the state machine diagram, where nominal is a central state and has transitions to DETUMBLING, LOW_POWER and PAYLOAD. Would that work?

s026134 (approved):

flight/apps/telemetry/packing.py
```diff
@@ -112,92 +114,92 @@ def pack_tm_frame(cls):
             eps_data = DH.get_latest_data("eps")
 
             # Mainboard voltage
-            cls._FRAME[17:19] = pack_signed_short_int(eps_data, EPS_IDX.MAINBOARD_VOLTAGE)
```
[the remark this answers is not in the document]
Ibrassow: We can find another way in a subsequent PR. Rn, this is one of the good uses of Copilot (changing the indices).

Akshat412 (approved):

flight/core/states.py
Ibrassow: Good catch for the transition back. 

I think we should keep the transition to DETUMBLING <-> LOW-POWER.

Ibrassow:
HW test before merge

Akshat412 (changes requested):
image
Commanding module seems to have a small bug. FSW log running on the v2 mainboard attached for reference (error message on log for [COMMAND] task).

Akshat412 (approved):
bugs fixed, godspeed

PR #86, opened 2025-01-17
Cpp sim interface

kmkar31:
- Replaces argusloop with the C++ sim for SIL testing
- Verified all sensors and actuators work
- Currently, there is a bug in ADCS FSW (indexing issues) that causes it to think gyro reading is always 0 (only for ACS). So it never switches out of detumbling. These fixes are in a separate branch (adcs-bug-fixes)

- Submodule currently points to a branch of GNC-sim because main is broken and needs quite some time to fix
- Sim Installed and Runs on the lab server


emulator/simulator.py
```diff
@@ -6,207 +6,82 @@
...
-        }
-
-    def gyro(self) -> np.ndarray:
-        """
-        Measures the gyroscope data.
+        # Initialize the CPP sim
+        trial = random.randint(0, 100)
+        RESULTS_FOLDER = os.path.join(RESULTS_ROOT_FOLDER, datetime.now().strftime("%Y-%m-%d_%H-%M-%S"))
+        os.mkdir(RESULTS_FOLDER)
+        self.cppsim = cppSim(trial, RESULTS_FOLDER, CONFIG_FILE)
```
Ibrassow, line 32: I don't like seeing this randomization at this level. I think we need a proper way to configure the simulator (spoiler: config files) but I will let it slide for the sake of progress (+ not critical for now)

Ibrassow (approved):

Akshat412:
Nothing looks off here, but can I see the new SIL integration into the CI in action before we merge into main?

Akshat412 (approved):
Nice
image

atletea (approved):
Tried with the SIL CI stuff I had and works well, should be good to merge

PR #92, opened 2025-01-20
Reading IMU uncalibrated data

perrintong325:


flight/hal/argus_v2.py
Akshat412, line 1: Exceptions should only be printed in the debug condition

flight/hal/argus_v2.py
```diff
@@ -441,22 +441,19 @@ def __imu_boot(self) -> list[int]:
...
+            imu.enable_feature(BNO_REPORT_UNCAL_MAGNETOMETER)
+            imu.enable_feature(BNO_REPORT_UNCAL_GYROSCOPE)
             if self.__middleware_enabled:
                 imu = Middleware(imu)
 
             self.__imu = imu
             self.__imu_temp_flag = False
             self.__device_list.append(imu)
         except Exception as e:
+            print(e)
```
Akshat412, line 456: Exceptions should only be printed in the debug conditional

Akshat412 (changes requested):
Double check that the changes to the HAL are also reflected in the emulator. Test with emulator / simulator.

perrintong325:
HAL interface unchanged, tested on HW

Akshat412 (approved):

PR #97, opened 2025-01-22
Removed middleware from flight and emulator, updated firmware

perrintong325:
Tested on HW and emulator, no change to HAL


Ibrassow (approved):
LGTM

PR #102, opened 2025-01-23
ADCS task implementation

kmkar31:
Implements the following:
- ADCS task logic dependeing on state
- MEKF implementation for sun sensor, magnetometer and gyro
    - Does initialization as well
 - Removes IMU task as a separate task
 - Implements Orbit propagator (without J2 effects)

TODO:
- Sensor validity checks based on review comments
- Decision on how to propagate magnetometer measurements if IMU fails in nominal
- Integrating ACS into architecture (will be done in another PR with MCM fixes)
- Addition to unit tests

Testing
- Tested on ARGUSv1.1 with spoofed GPS, sun sensor and mag measurements


flight/apps/adcs/ad.py
```diff
@@ -9,52 +9,423 @@
...
+        - This function is not directly written into init to allow multiple retires of initialization
+        - Sets the initialized attribute of the class once done
+        """
+        # Get a valid GPS position
+        gps_valid, gps_record_time, gps_pos_ecef, gps_vel_ecef = self.read_gps()
+
+        # Spoofing GPS
+        gps_pos_ecef = np.array([-1325.226552563400, 4849.749546663410, -4575.122922351300]) * 1000
+        gps_vel_ecef = np.array([-5.48091694803860, -4.38278368616351, -3.05761643533641]) * 1000
+        gps_valid = 1
```
Ibrassow, line 149: Remove the spoof here, just have the emulator (GPS) return this fixed measurement (the sim returns its own values anyway). No hardcode in flight, even for spoofed data

flight/apps/adcs/ad.py
```diff
@@ -9,52 +9,423 @@
...
+
+        # Get a valid sun position
+        sun_status, sun_pos_body, lux_readings = self.read_sun_position()
+
+        # Spoofing Sun Status
+        sun_status = SUN_VECTOR_STATUS.UNIQUE_DETERMINATION
+        sun_pos_body = np.array([0.93632918, 0.35112344, 0.0])
+
+        if sun_status == SUN_VECTOR_STATUS.NO_READINGS or sun_status == SUN_VECTOR_STATUS.NOT_ENOUGH_READINGS:
+            print("Failed Sun Status")
```
Ibrassow, line 171: logger instead of print

flight/apps/adcs/ad.py
```diff
@@ -9,52 +9,423 @@
...
+        gps_valid, gps_record_time, gps_pos_ecef, gps_vel_ecef = self.read_gps()
+
+        # Spoofing GPS
+        gps_pos_ecef = np.array([-1325.226552563400, 4849.749546663410, -4575.122922351300]) * 1000
+        gps_vel_ecef = np.array([-5.48091694803860, -4.38278368616351, -3.05761643533641]) * 1000
+        gps_valid = 1
+        gps_record_time = int(time.time())
+
+        if not gps_valid:
+            print("Failed GPS")
```
Ibrassow, line 153: logger

flight/apps/adcs/ad.py
```diff
@@ -9,52 +9,423 @@
...
+            gps_pos_eci = np.dot(R_ecef2eci, gps_pos_ecef)
+            gps_vel_eci = np.dot(R_ecef2eci, gps_vel_ecef)
+            true_pos_eci, true_vel_eci = propagate_orbit(current_time, gps_record_time, gps_pos_eci, gps_vel_eci)
+
+        # Get a valid sun position
+        sun_status, sun_pos_body, lux_readings = self.read_sun_position()
+
+        # Spoofing Sun Status
+        sun_status = SUN_VECTOR_STATUS.UNIQUE_DETERMINATION
+        sun_pos_body = np.array([0.93632918, 0.35112344, 0.0])
```
Ibrassow, line 168: Same as above, no spoof, have those fixed values generated through the emulator

flight/apps/adcs/ad.py
```diff
@@ -9,52 +9,423 @@
...
 from ulab import numpy as np
 
+"""
+    Attitude Determination Class
+    Contains functions for the following:
+        - Initializing the MEKF
+        - Reading Sensors
+        - Propagating MEKF
+"""
+
```
Ibrassow, line 31: For the whole file: 
- for the sake of clarity (especially for other people reading the code), I would simply define the boolean flag that your return for your status (so the code is self-documenting) ~ instead of returning 1 or 0, return STATUS_SUCCESS or STATUS_FAILED (or whatever constant name) that you can use for all your functions.
- Given Python's permissive nature, I'd add clear definitions of the expected types for args and return types in your docstrings (and add type annotations)

For flight code, you want to document as much as possible (it's also a self.check)

flight/tasks/adcs.py
```diff
@@ -97,108 +86,127 @@ async def main_task(self):
...
+                # Query Magnetometer
+                self.AD.magnetometer_update(self.time, update_covariance=False)
+
+                # Run Attitude Control
+                self.attitude_control()
+
+                # Log Data
+                # NOTE : Most of these values will be 0 since MEKF is not initialized
+                self.log()
+
```
Ibrassow, line 106: We should be logging every time, except in STARTUP mode. Put it outside that state logic. Since you have the write_interval arguments in the DataHandler, it will not write to the SD card every time.

calling this function multiple times is cheap in terms of memory usage because Python operates on a reference-counting system and the assignment in your task log() a(and also in DataHandler) do not create deep copies

Ibrassow: @Akshat412 and @s026134  you might need that information too

flight/tasks/adcs.py
```diff
@@ -97,108 +86,127 @@ async def main_task(self):
...
+                else:
+                    if self.execution_counter == 0:
+                        # TODO : Turn coils off before measurements to allow time for coils to settle
+                        self.execution_counter += 1
+                        return
+
+                    elif self.execution_counter < 4:
+                        self.execution_counter += 1
+                        return  # do nothing
```
Ibrassow, line 148: In the same line, don't done early returns main_task (+ you need to log at the end)

flight/tasks/adcs.py
```diff
@@ -97,108 +86,127 @@ async def main_task(self):
...
-            elif not self.eclipse_state and not np.linalg.norm(sun_vector_err) < MCMConst.POINTING_ERROR_TOL:
-                dipole_moment = get_sun_pointing_dipole_moment(
-                    imu_mag_data,
-                    pointing_err,
-                )
-            else:
-                dipole_moment = np.zeros(3)
-            MagneticCoilAllocator.set_voltages(dipole_moment)"""
+                # Query the Gyro
+                self.AD.gyro_update(self.time, update_covariance=False)
```
Ibrassow, line 95: For file logging purposes, I think you want to dissociate the frequency of gyro reading and the actual AD updates (?)

flight/tasks/adcs.py
```diff
@@ -97,108 +86,127 @@ async def main_task(self):
...
+            # LOW POWER
+            # ------------------------------------------------------------------------------------------------------------------------------------
+            elif SM.current_state == STATES.LOW_POWER:
+
+                # In Low power, it is assumed that the satellite is about to die.
+                # In that case, the satellite will have to re-initialize the MEKF again
+
+                # No AD or ACS runs in LOW-POWER
+                if self.AD.initialized:
+                    self.AD.initialized = False
```
Ibrassow, line 122: Ok for ACS. Why would you disable the AD? querying the IMU is cheap (and MEKF running isn't what will kill us). The CPU still runs at the same frequency anyway. 

Also, in LOW_POWER, you want to know your physical state and get IMU measurements. The AD must only be conditioned by the tumbling state, not power. So you can seamlessly continue the AD if you came from nominal, and continue all the rest of the processing and logging

Ibrassow (changes requested):

flight/apps/adcs/ad.py
kmkar31: Yep, apps aren't going to use logging/print. Akshat wanted to run this on ARGUSv2. I've kept these in so he can verify it runs on his board. After that, I'll add a clean up commit

Akshat412: Yeah, this is still somewhat unverified on working hardware (Karthik's board does not have a working IMU). After that I think a lot of these debug statements either need to be made logger warnings or (if not severe enough errors to report in a log) just disabled.

flight/apps/adcs/ad.py
Akshat412: Yeah we'd want these to be vectors of 0s in actual FSW applications (that is the default for the GPS module if it gets no fix anyways). I think this was added today just for some feature testing, but needs to be removed before review.

flight/tasks/adcs.py
Akshat412: Yeah, I was thinking state transitions for detumbling can just be based on the ADCS.MODE similar to the EPS.STATE_FLAG. The state machine implementation should just take these two flags as inputs and be the centralized entity for state transitions.

flight/apps/adcs/ad.py
Ibrassow: You can use the logger.warning here. We'll integrate into the CI tools that will analyze our logging. Do make use of the serial logging.

Right now, you can disable all the serial logging by setting the level to NOTHING (which will be used for flight)

flight/tasks/adcs.py
Ibrassow: I just pushed the query of the previous state. For some reasons, the logic was there but not the interface to access it. 
image
No need for complicated state logic

flight/apps/adcs/ad.py
kmkar31: Wait, I thought Akshat did not want logging from apps and they should only go thru tasks

flight/apps/adcs/ad.py
kmkar31: Yep, I'll add these to the emulator

flight/tasks/adcs.py
kmkar31: Wait, what is the point of this previous state. Akshat and I discussed that we don't want adcs task directly switching the global state, but I don't see previous states do here

flight/tasks/adcs.py
kmkar31: I'm not sure I understand. We can read the gyro faster than we are currently (i.e., the remaining 4 steps where we do nothing can also be used to query gyro) but I suspect we can't afford it from a timing perspective. But how does any of this affect logging?

flight/tasks/adcs.py
kmkar31: I assume in low power, we die. Might as well kill all ADCS and re-initialize after we boot again. In Low power, we get rid of GPS anyway (I assume), so our orbit propagation will drift significantly over ~ 1 day. So we would be better of re-initializing the MEKF anyhow.

flight/apps/adcs/ad.py
Ibrassow: You can log stuff anywhere you want

kmkar31: Cool, if it isn't a standard we set, I'll change all of these into logging

flight/tasks/adcs.py
Ibrassow: It allows you to know if you continue the AD or not by querying the previous (if you were coming from NOMINAL or DETUMBLING). But since you have the initialized flag for AD, it doesn't really matter for you. 

Anyway, the function needed to be there irrespective of your PR.

flight/tasks/adcs.py
Ibrassow: It just seems that you are reading the gyro at 5Hz only in detumbling and 1Hz in Nominal/XP. I'd simply get the IMU reading at a fixed frequency (irrespective of state since it doesn't matter) 

~~

if you don't read the gyro, your logging won't get an updated value (only the previous one that was already there).

I wouldn't assume any timing stuff until tested on the new MCU, just do what needs to be done, we'll optimize when we need to.

flight/tasks/adcs.py
Ibrassow: LOW_POWER == pre-emptive energy saving mode, nothing to do with us dying. We can totally run the GPS at a lower rate in LOW_POWER (even updates every X min is valuable). There is no point in killing the AD. We will know on the ground that our GPS runs at a lower frequency or is off.

Depending on our orbit propagation, the errors won't be as big as you think. Our dynamics is kind

flight/apps/adcs/ad.py
kmkar31: Done. See commits : cacba2f998ee126a228c7fe11e43e1f358907eca and c528027e7ae53f736d71862d33bff624553a71bc

flight/tasks/adcs.py
kmkar31: Done. See https://github.com/cmu-argus-2/FSW-mainboard/pull/102/commits/9bdb0de4dc9d75f8d2f8fccaa177f58afd78ade3

flight/tasks/adcs.py
kmkar31: Done. See https://github.com/cmu-argus-2/FSW-mainboard/pull/102/commits/9bdb0de4dc9d75f8d2f8fccaa177f58afd78ade3

flight/apps/adcs/ad.py
kmkar31: Done. See https://github.com/cmu-argus-2/FSW-mainboard/pull/102/commits/345deca219d7f3ebea0f74e79894552cbaa29ebf

flight/apps/adcs/ad.py
kmkar31: Done. See https://github.com/cmu-argus-2/FSW-mainboard/pull/102/commits/345deca219d7f3ebea0f74e79894552cbaa29ebf

flight/tasks/adcs.py
kmkar31: Done. See https://github.com/cmu-argus-2/FSW-mainboard/pull/102/commits/e480cf79c7b39a47f6803ae65c8bfab5fdcac7ce

flight/tasks/adcs.py
kmkar31: Done. See https://github.com/cmu-argus-2/FSW-mainboard/pull/102/commits/a792476f12f2ca889e4ae7ed8d17f5451dec67a0

flight/tasks/adcs.py
```diff
@@ -2,33 +2,23 @@
...
 
+"""
+    ASSUMPTIONS :
+        - ADCS Task runs at 5 Hz (TBD if we can't handle this)
+        - In detumbling, control loop executes every 200ms
+        - In nominal/experiment, for 4 executions, we do nothing. On the fifth, we run full MEKF + control
+"""
+
 
 class Task(TemplateTask):
```
kmkar31, line 23: I want to remove 'eclipse' as a data field from logging. From an ADCS PoV, there is no difference between staying in eclipse and fried sensors. If we definitely want an eclipse flag, I could add it sun status

kmkar31:
Fixed all suggested changes.
- Found a math bug that somehow did not get flagged on hardware
- Scaling helps with $S^{-1}$ and has been added (This still has a try except because ulab has a lot of conditions to flag singularity and I can't if check them all)
- Emulator works without errors as well

flight/tasks/adcs.py
```diff
@@ -97,108 +85,139 @@ async def main_task(self):
...
-    return x != x
+                if not self.AD.initialized:
+                    self.AD.initialize_mekf()
+
+                elif (
+                    SM.current_state == STATES.NOMINAL
+                    and np.linalg.norm(self.AD.state[self.AD.omega_idx]) > ModeConst.EKF_INIT_TOL
+                    and not DH.get_latest_data("cdh")[CDH_IDX.DETUMBLING_ERROR_FLAG]
+                ):
+                    SM.switch_to(STATES.DETUMBLING)
```
Ibrassow, line 148: Change to local state state only. General SM changes handled by CDH

Ibrassow (changes requested):
Approval pending hardware testing (and subsequent removal of print statement) and state switch handled locally only.

flight/tasks/adcs.py
kmkar31: Added eclipse flag into sun status. It now takes precedence over everything except no readings.

flight/tasks/adcs.py
```diff
@@ -97,108 +86,127 @@ async def main_task(self):
...
+
+                # Log Data
+                # NOTE : Most of these values will be 0 since MEKF is not initialized
+                self.log()
+
+                # Check if detumbling has been completed
+                if np.linalg.norm(self.AD.state[self.AD.omega_idx]) <= ModeConst.STABLE_TOL:
+                    self.AD.initialize_mekf()
+                    SM.switch_to(STATES.NOMINAL)
+
```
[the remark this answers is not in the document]
kmkar31: Done. ADCS MODE used to change state

Akshat412:
Works on hardware! 

But, RAM use is up to 94% now. Not sure how much of this can be optimized.

Screenshot from 2025-02-08 14-42-55

Akshat412:
@kmkar31 @Ibrassow I am making some changes to the TM packing since we got rid of the ECLIPSE field. This PR looks 99% ok to me, my only ask would be a reduction in the amount of log data printed. Since ADCS runs at 5 Hz the more log data you have the slower we get. 

Obviously for flight this does not matter (logging will be off) but even for development, I think we can get away with less text output per task run.

kmkar31:
> @kmkar31 @Ibrassow I am making some changes to the TM packing since we got rid of the ECLIPSE field. This PR looks 99% ok to me, my only ask would be a reduction in the amount of log data printed. Since ADCS runs at 5 Hz the more log data you have the slower we get.
> 
> Obviously for flight this does not matter (logging will be off) but even for development, I think we can get away with less text output per task run.

Since I have an execution counter, I can modify to log at 1Hz instead of 5Hz. That'll save some time. I'd also rather do RAM improvements in another PR. I want to integrate ACS (Derek's stuff) in as well before optimizing

Ibrassow (approved):

Akshat412 (approved):

PR #106, opened 2025-01-24
Max17205 driver integration

achstar:


flight/tasks/eps.py
```diff
@@ -74,6 +74,23 @@ def read_vc(self, sensor, voltage_idx, current_idx):
...
+        # read values from MAX17205
+        fuel_gauge = SATELLITE.FUEL_GAUGE
+        if (fuel_gauge is not None):
+            self.log_data[EPS_IDX.BATTERY_PACK_REPORTED_SOC] = int(fuel_gauge.read_soc())
+            self.log_data[EPS_IDX.BATTERY_PACK_REPORTED_CAPACITY] = int(fuel_gauge.read_capacity())
+            self.log_data[EPS_IDX.BATTERY_PACK_CURRENT] = int(fuel_gauge.read_current())
+            self.log_data[EPS_IDX.BATTERY_PACK_VOLTAGE] = int(fuel_gauge.read_voltage())
+            self.log_data[EPS_IDX.BATTERY_PACK_MIDPOINT_VOLTAGE] = int(fuel_gauge.read_midvoltage())
+            self.log_data[EPS_IDX.BATTERY_CYCLES] = int(fuel_gauge.read_cycles())
+            self.log_data[EPS_IDX.BATTERY_PACK_TTE] = int(fuel_gauge.read_tte())
```
Akshat412, line 87: Update the telemetry packing to reflect the changes in size for TTE, TTF, and remove cycles and time since power up. The indexes should not change because of these changes, if they do just leave it for later.

Akshat412 (changes requested):
Just the change for the TM packing, everything else looks good!

flight/tasks/eps.py
achstar: Sounds good. I don't think the cycles and TimerH (what's currently labeled as time since power up) registers are that useful, but they do provide some information on cell chemistry. I think the Age register might be more useful in that regard, though, since that provides a percentage of present cell capacity to expected cell capacity.

Ibrassow (approved):
LGTM

flight/tasks/eps.py
Akshat412: I'm Ok either way, whatever you feel is best. I checked with Neil and he agreed that they aren't strictly important, but we still have enough space in the nominal frame to downlink them. We also plan on having a more specific hardware / HAL frame, so these could be part of that.

achstar: Made the above changes and tested it on hardware, looks good 👍

Akshat412 (approved):
LGTM

PR #110, opened 2025-01-26
Mcm tests

Herpderk:
Removed random input generation, added nominal (zero-error), and zero (parallel argument vector) test cases. Also removed incompatible libraries and changed some parameters for SIL (including inertia matrix).


tests/test_mcm.py
```diff
@@ -20,35 +22,43 @@
...
-    dipole_moment = np.random.uniform(size=3)
+def inputs() -> dict:
+    return {
+        "magnetic_field": np.array([1.0, 0.0, 0.0]),
+        "sun_vector": np.array([1.0, 0.0, 0.0]),
+        "spin_error": np.zeros(3),
+        "pointing_error": np.zeros(3),
+        "dipole_moment": np.zeros(3),
+    }
+
```
Ibrassow, line 33: think there needs to have also other inputs (that are wrong like an array of size 2) and check that it throws an error effectively (instead of implicit computations that are common with Python)

tests/test_mcm.py
```diff
@@ -61,15 +71,39 @@ def test_controller_handler() -> None:
...
+        "spin_error": np.array([1.0, 0.0, 0.0]),
+        "pointing_error": np.array([1.0, 0.0, 0.0]),
+    }
+
+
+def test_spin_stabilizing_near_zero(parallel_inputs) -> None:
+    # Check for NaN, which may occur when the argument vectors are parallel
+    dipole_moment = get_spin_stabilizing_dipole_moment(parallel_inputs["magnetic_field"], parallel_inputs["spin_error"])
+    assert dipole_moment.shape == (3,)
+    assert not np.any(np.isnan(dipole_moment))
```
Ibrassow, line 102: Q: what is supposed to output if not NaN in this case?

flight/apps/adcs/mcm.py
```diff
@@ -92,7 +91,7 @@ class MagneticCoilAllocator:
         "ZM": 0.0,
     }
 
-    mat = copy.deepcopy(MCMConst.ALLOC_MAT)
+    mat = MCMConst.ALLOC_MAT[:]
```
Ibrassow, line 94: n my recollection, this is not the only CP conflict, this needs to be tested on hardware

Ibrassow (changes requested):
Couple of things to fix:
- Conflicts must be resolved (pull from main)
- ABSOLUTELY need to be tested on hardware to confirm the conflicts with CP are fixed

tests/test_mcm.py
Herpderk: I think the output would be zero, but I was hesitant to assert that because in a "black box" scenario, the output could be any defined number

flight/apps/adcs/mcm.py
Herpderk: Gotcha

tests/test_mcm.py
Herpderk: Do you want to add assertions in the flight code?

Herpderk:
TODOs:
1. momentum_target matmul doesn't work
2. cast all ADCS data as arrays

Ibrassow:
> TODOs:
> 
>     1. matrix multiplication operator (`@`) doesn't work for ulab `ndarray`'s.
> 
>     2. cast all ADCS data as arrays.
> 
>     3. Fix `ControllerHandler` spin-axis assignment
> 
> 
> @Ibrassow any ideas for 1 (besides implementing `matmul` in python)

You best bet is to implement it (in an efficient way). Ideally, we add it directly to the C firmware

flight/apps/adcs/mcm.py
```diff
@@ -55,17 +56,16 @@ class ControllerHandler:
     _u_max_convert = MagnetorquerConst.V_MAX / MagnetorquerConst.V_CONVERT
 
     # Init spin direction
-    _eigvals, _eigvecs = np.linalg.eig(PhysicalConst.INERTIA_MAT)
+    # IMPORTANT: ulab's eig returns (eigenvectors, eigenvalues)
+    _eigvecs, _eigvals = np.linalg.eig(PhysicalConst.INERTIA_MAT)
```
Herpderk, line 60: @Ibrassow Numpy's `eig` function returns (eigenvalues, eigenvectors), but in ulab it's the reverse order. This causes FSW on the mainboard and the unit tests to be incompatible. Is there a flag we can pass into MCM to let it know when it's being run in SIL vs. on the mainboard?

tests/test_mcm.py
Herpderk: Circling back to this, if both arguments for a given control law are not 3D but are of  the same dimension, the  computation will not raise an error. The only way to enforce 3D arguments is add an assertion, but I'm not sure if that's worthwhile for flight software.

Ibrassow:
Closing since outdated

PR #130, opened 2025-01-31
Eps state flag

achstar:
Adding flag to EPS in order to indicate when to enter/exit the low power state, nominal state, and experiment/payload state


flight/tasks/eps.py
```diff
@@ -8,6 +8,11 @@
 from core import state_manager as SM
 from core.states import STATES
 from hal.configuration import SATELLITE
+from micropython import const
+
+EPS_STATE_NOMINAL = const(0x0)
+EPS_STATE_LOW_POWER = const(0x1)
+EPS_STATE_PAYLOAD = const(0x2)
```
achstar, line 16: Is there a better place to put these?

flight/tasks/eps.py
```diff
@@ -56,9 +61,21 @@ class Task(TemplateTask):
...
+    LOW_POWER_ENTRY_SOC_THRESHOLD = 30
+    LOW_POWER_EXIT_SOC_THRESHOLD = 40
+    PAYLOAD_ENTRY_SOC_THRESHOLD = 80
+    PAYLOAD_EXIT_SOC_THRESHOLD = 60
+
+    eps_state = {
+        EPS_STATE_LOW_POWER : "LOW POWER",
+        EPS_STATE_NOMINAL : "NOMINAL",
+        EPS_STATE_PAYLOAD : "PAYLOAD",
+    }
```
achstar, line 78: Can potentially remove these if they take up too much memory, they make things a bit more readable

flight/apps/telemetry/packing.py
```diff
@@ -328,11 +330,11 @@ def pack_tm_heartbeat(cls):
...
-                cls._FRAME[227:229] = pack_unsigned_short_int(thermal_data, THERMAL_IDX.IMU_TEMPERATURE)
+                cls._FRAME[228:230] = pack_unsigned_short_int(thermal_data, THERMAL_IDX.IMU_TEMPERATURE)
                 # CPU temperature
-                cls._FRAME[229:231] = pack_unsigned_short_int(thermal_data, THERMAL_IDX.CPU_TEMPERATURE)
+                cls._FRAME[230:232] = pack_unsigned_short_int(thermal_data, THERMAL_IDX.CPU_TEMPERATURE)
                 # Battery temperature
-                cls._FRAME[231:233] = pack_unsigned_short_int(thermal_data, THERMAL_IDX.BATTERY_PACK_TEMPERATURE)
+                cls._FRAME[232:234] = pack_unsigned_short_int(thermal_data, THERMAL_IDX.BATTERY_PACK_TEMPERATURE)
             else:
                 logger.warning("No latest Thermal data available")
```
achstar, line 339: TODO: make corresponding change in spreadsheet when this PR is finalized

flight/tasks/eps.py
```diff
@@ -129,4 +146,34 @@ async def main_task(self):
...
+                elif ((soc >= self.PAYLOAD_EXIT_SOC_THRESHOLD)
+                      and (soc <= self.PAYLOAD_ENTRY_SOC_THRESHOLD)):
+
+                    if (SM.current_state == STATES.EXPERIMENT):
+                        self.log_data[EPS_IDX.EPS_STATE] = EPS_STATE_PAYLOAD
+                    else:
+                        self.log_data[EPS_IDX.EPS_STATE] = EPS_STATE_NOMINAL
+
+                else:  # greater than payload entry threshold
+                    self.log_data[EPS_IDX.EPS_STATE] = EPS_STATE_PAYLOAD
```
Ibrassow, line 175: has the whole logic been tested? I'd put that inside a function "whatener_name(soc)" and return the EPS_STATE_X instead (in the mentioned eps file (see above)). Then you can:
1) add unit tests for that specific function (with this PR)
2) in the task, you call that function and do the log just after
(or even shorter: self.log_data[EPS_IDX.EPS_STATE] = you_function(soc)) 
This would be much clearer and more reliable  

Do make sure to make the unit tests on this PR

Ibrassow (changes requested):

flight/apps/eps/eps.py
```diff
@@ -0,0 +1,51 @@
...
+        flag = EPS_POWER_FLAG.LOW_POWER
+
+    elif ((soc > EPS_SOC_THRESHOLD.LOW_POWER_ENTRY)
+            and (soc < EPS_SOC_THRESHOLD.LOW_POWER_EXIT)):
+
+        if (curr_flag == EPS_POWER_FLAG.LOW_POWER):
+            flag = EPS_POWER_FLAG.LOW_POWER
+        else:
+            flag = EPS_POWER_FLAG.NOMINAL
+
```
achstar, line 32: leaving the below logic untested for now because i expect it to change, keeping a very simplified approach (i.e. just using SOC threshold to determine when to enter/exit experiment state from a power monitoring standpoint)

flight/apps/telemetry/packing.py
```diff
@@ -115,89 +115,91 @@ def pack_tm_heartbeat(cls):
             eps_data = DH.get_latest_data("eps")
 
             if eps_data:
```
achstar, line 117: Observed the following error when running emulate: [EPS] pack expected 42 items for packing (got 41)
I didn't check if I saw this error after the previous EPS change, but does anyone have any idea of what might be causing it? I didn't see it when testing on hardware previously

tests/test_eps.py
achstar, line 1: Anything to add to this test? Did not add any pytest features because they didn't seem necessary

achstar:
Not yet tested on hardware

tests/test_eps.py
```diff
@@ -0,0 +1,40 @@
...
+
+    # Low power exit range and above
+    soc = random.randint(EPS_SOC_THRESHOLD.LOW_POWER_EXIT, 100)
+
+    flag = GET_EPS_POWER_FLAG(EPS_POWER_FLAG.NONE, soc)
+    assert flag != EPS_POWER_FLAG.LOW_POWER
+    flag = GET_EPS_POWER_FLAG(EPS_POWER_FLAG.LOW_POWER, soc)
+    assert flag != EPS_POWER_FLAG.LOW_POWER
+    flag = GET_EPS_POWER_FLAG(EPS_POWER_FLAG.NOMINAL, soc)
+    assert flag != EPS_POWER_FLAG.LOW_POWER
```
Ibrassow, line 40: In general, you shouldn't use random values in unit tests (just take one fixed value in your specific range) as they need to be deterministic and thus replicable through runs. Lots of hardcoding but that's the way :)

tests/test_eps.py
Ibrassow: since this is the main addition, you only need to test that function. 

Edge cases should be tested as well. what happens if you get a negative soc like -2 ? or > 100?
or if the curr_flag is outside the range of accepted constants? 

You'll see that you either decide to raise an exception (which should be documented and tested here) or you return a particular value (your NONE state might then be more useful, as an UNDEFINED)

Ibrassow (changes requested):

tests/test_eps.py
Ibrassow: Not sure if this was caused by the random generator but the unit test fails on my machine 
image

achstar: Ah yeah I set the random range incorrectly, it's inclusive of the low power exit threshold when it shouldn't be.
And makes sense, random testing doesn't make much sense for this test. I'll add a bunch of hardcoded test cases

tests/test_eps.py
```diff
@@ -1,40 +1,63 @@
...
 
-    # Low power SOC range
-    soc = random.randint(0, EPS_SOC_THRESHOLD.LOW_POWER_ENTRY)
+    # Invalid SOC values
+    socs = [-2, 101]
+    for soc in socs:
+        flag = GET_EPS_POWER_FLAG(EPS_POWER_FLAG.NONE, soc)
+        assert flag == EPS_POWER_FLAG.NONE
+        flag = GET_EPS_POWER_FLAG(EPS_POWER_FLAG.LOW_POWER, soc)
+        assert flag == EPS_POWER_FLAG.NONE
```
Ibrassow, line 13: given the amount of input/outputs you're testing, using @pytest.mark_parameterize (see pic below) here might make everything clearer. (You can do it now or another PR later (up to you)
image

Ibrassow (commented):
Pending hardware testing, otherwise it looks good

tests/test_eps.py
achstar: Hardware testing done so PR should be good to go. Also checked that emulate works

Ibrassow (approved):

flight/apps/telemetry/packing.py
Akshat412: RAM usage at 90% is extremely bad but I don't think this byte is contributing much to it. It's been around 90 for a while now.

We should do a review of what in FSW is adding so much RAM usage (HAL is about 60% currently) and optimize the modules that are culprits. The 2350 also has double the RAM which should help out with this.

Akshat412 (approved):
Everything looks good to me! Do remember to update the spreadsheet, thanks

PR #140, opened 2025-02-03
Integrated file ID and file time checks into file transfer requests

Akshat412:
Comms FSW reworked to match new command definitions for GS file requests.

### DIFF
--- a/flight/apps/comms/comms.py
+++ b/flight/apps/comms/comms.py
@@ -262,7 +262,6 @@ class SATELLITE_RADIO:
         else:
             # Valid filepath from DH, set size and message count
             file_stat = os.stat(cls.filepath)
-            print(file_stat)
 
             # TODO: Associate file IDs with subsystems
             cls.file_ID = 0x01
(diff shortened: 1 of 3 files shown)


flight/apps/comms/comms.py
```diff
@@ -346,16 +359,55 @@ def transmit_file_metadata(cls):
...
+    def check_rq_file_params(cls, packet):
+        """
+        If file MD requested, use this to see if
+        new filepath is needed, and RQ CDH for one.
+
+        If file PKT requested and params do not match,
+        this is used for error detection to default to HB state.
+
+        For now, assume that hardcoded filepath
+        matches the file the GS requested
```
Ibrassow, line 388: Just added the fct to extract the time form the file path to your PR, so should solve this

Ibrassow (approved):
All good for me (+ good since it's been tested). Up to you but you could add directly the correct logic to handle the filename checks

flight/apps/comms/comms.py
Akshat412: Thanks! In that case (note to other reviewers too) let me test the new functionality on hardware before approving the merge.

Akshat412:
Tested on hardware, everything works correctly. Also updated the CI versioning on my branch so a merge to main should fix that.

sa2415 (approved):

s026134 (approved):

PR #144, opened 2025-02-06
updated move to board script

perrintong325:
updated the build script so other teams don't create wrong directories and come complaining to me that FSW is broken


Ibrassow (approved):

Akshat412 (approved):
Hah. Nice.

PR #148, opened 2025-02-10
Attitude Control Stack integrated into ADCS

kmkar31:
PR achieves the following
- Integrates ACS into the existing ADCS task
- Runs Sensor validity checks before using them for processing
- Determines a const class for all logging
    - Logging now only happens at 1Hz
- Tested on Argus v1.1

TODO / Decisions to be made:
- Cannot actually test coil energizing without a flatsat

No bugs observed that would crash main but cannot confirm without X-Y boards


Ibrassow:
@kmkar31 AS we discussed,
1. ACS should be independent of the MEKF. MEKF just happens to improve the angular velocity state (through the bias)
2. Flatsat building is a terrible bottleneck. Make people move faster once the new boards arrive

Ibrassow (changes requested):
This looks generally good to me. Unfortunate we can't make yet the full hardware testing. @kmkar31 do raise an issue about it so someone can do it as soon as the new flatsat is there. 

Could you zero the coil voltage when we transition to LOW_POWER?

kmkar31:
Added
- ACS even if MEKF not running
- Coils zero out either for coils to settle or to conserve power

@Akshat412 pls verify code changes on ARGUS v2 again

Ibrassow (approved):
LGTM!

I let @Akshat412 do the the tests on HW v2

Akshat412:
The current ADCS stack doesn't play well with GPS data (NoneType issues and also somehow forever loops in the orbit propagator). Can fixes for this be made in the current PR or do you want that to be separated?

Ibrassow:
Better to fix here and use the simulator for the intermittent GPS fixes

Akshat412:
So, HIL with the GPS running seems to work ok. The only thing I can see breaking is this single error that happens sometimes in DETUMBLING.
image

kmkar31:
@Akshat412  fixed matmul error. Realized IMU driver returns mag and gyro data in tuple format. I did not cast them into numpy arrays which, when multiplied by a constant to convert into appropriate units, threw an error

Akshat412:
@kmkar31 sounds good, thanks! I will test this on the hardware tomorrow.

Akshat412:
@kmkar31 @Ibrassow FSW is running well in both DETUMBLING and NOMINAL. Should be good to merge.

PR #155, opened 2025-02-14
Comms Source and Destination Header

Akshat412:


flight/apps/telemetry/packing.py
Akshat412, line 1: If the radio ever tries transmitting more than 256 bytes we end up not transmitting anything. Given overheads in this due to RadioHead etc. it's best to stay far away from this maximum for robustness.

Ibrassow: We are hardcoding everything anyway so why waste precious bytes (esp in file transfer)?

Akshat412: Hmm, this is true, 240 bytes is too restrictive this can become 248. I'll change it.

s026134 (approved):

flight/apps/comms/comms.py
```diff
@@ -35,13 +35,20 @@ class MSG_ID:
     Comms message IDs that are downlinked during the mission
     """
 
+    # Source header ID for Argus - THIS MUST BE UNIQUE FOR EACH SPACECRAFT
+    ARGUS_ID = 0x01
+
+    # Source header ID for GS
+    GS_ID = 0x04
```
Ibrassow, line 42: I don't think we should hardcode the GS that sends the command. All GS should be able to receive and process all responses. See it like all GS have a "common state". Ultimately, sending a command might come from one GS (that we decide per command process) but the RX side is handled by whoever processes it first / or any other rule (we'll develop later the handling for this)

flight/apps/comms/comms.py
```diff
@@ -84,7 +91,7 @@ class SATELLITE_RADIO:
     state = COMMS_STATE.TX_HEARTBEAT
 
     # Init TM frame for preallocating memory
-    tm_frame = bytearray(250)
+    tm_frame = bytearray(240)
```
Ibrassow, line 94: See comment

Ibrassow (changes requested):

flight/apps/comms/comms.py
```diff
@@ -35,13 +35,38 @@ class MSG_ID:
...
+    came from [src] or is going to [dst] a ground station. This
+    does not specify a particular GS, only that the packet is
+    associated with a downlink / uplink.
+    """
+
+    # Header ID for Argus - THIS MUST BE UNIQUE FOR EACH SPACECRAFT
+    ARGUS_ID = 0x01
+
+    # Header ID for all ground stations (genericized)
+    GS_ID = 0x04
```
Ibrassow, line 60: No 0xA1, 0xA2, 0xEA ? :(

Ibrassow (approved):

flight/apps/comms/comms.py
Ibrassow: you're not a fun person

PR #156, opened 2025-02-16
ADCS: Sun acquisition

kmkar31:
Waiting for #235 to be merged in (to test on HW with new logic)

- Adds 4 pyramid sensors to the existing 5 light sensors
- Adds ECLIPSE flag as a part of the sun status
- Modifies test cases to account for 9 light sensors


flight/apps/adcs/sun.py
```diff
@@ -82,21 +80,28 @@ def compute_body_sun_vector_from_lux(I_vec):
...
-
-    sun_body[0] = i_vec[0] if i_vec[0] > i_vec[1] else -i_vec[1]
-    sun_body[1] = i_vec[2] if i_vec[2] > i_vec[3] else -i_vec[3]
-    sun_body[2] = i_vec[4]
+    # Extract body vectors and lux readings where the sensor readings are valid
+    valid_sensor_idxs = [
+        idx for idx in range(len(I_vec)) if I_vec[idx] != ERROR_LUX and I_vec[idx] >= THRESHOLD_ILLUMINATION_LUX
+    ]
+    N_valid = PhysicalConst.LIGHT_SENSOR_NORMALS[valid_sensor_idxs, :]
+    I_valid = [I_vec[idx] for idx in valid_sensor_idxs]
```
Ibrassow, line 91: Prefer pre-allocating everything and literally hardcoding things (if we need hundreds if statement so be it). We should avoid any dynamic allocation, that includes list comprehension

flight/apps/adcs/sun.py
```diff
@@ -14,15 +14,16 @@
...
 from core import logger
 from hal.configuration import SATELLITE
 from micropython import const
 from ulab import numpy as np
 
 MAX_RANGE = const(117000)  # OPT4001
 THRESHOLD_ILLUMINATION_LUX = const(3000)
-NUM_LIGHT_SENSORS = const(5)
+NUM_LIGHT_SENSORS = const(9)
 ERROR_LUX = const(-1)
```
Ibrassow, line 27: Add leading underscore to all const

flight/apps/adcs/math.py
```diff
@@ -76,3 +76,38 @@ def quaternion_multiply(q1: np.ndarray, q2: np.ndarray) -> np.ndarray:
...
+
+    # Calculate cofactors (matrix of minors with alternating signs)
+    # Cofactor matrix (not yet transposed):
+    adjugate = np.array(
+        [
+            [(e * i - f * f), -(b * i - f * c), (b * f - e * c)],
+            [-(b * i - c * f), (a * i - c * c), -(a * f - b * c)],
+            [(b * f - c * e), -(a * f - c * b), (a * e - b * b)],
+        ]
+    )
```
Ibrassow, line 109: This matrix can be preallocated as a static variable in the file and be filled component by component instead

flight/apps/adcs/math.py
```diff
@@ -76,3 +76,38 @@ def quaternion_multiply(q1: np.ndarray, q2: np.ndarray) -> np.ndarray:
...
+
+    # Multiply adjugate by 1/det to get the inverse
+    inv = adjugate / det
```
Ibrassow, line 112: and here, this would do (excluding the operators' overload calls) 9 divisions. You can precompute once 1/det and then multiply all your terms, either at the end of filling adjugate or multiply each terms as you're filling them

Ibrassow:
Was it tested on HW?

flight/apps/adcs/math.py
kmkar31: Removed this function

flight/apps/adcs/math.py
kmkar31: Removed this function

flight/apps/adcs/sun.py
kmkar31: Done

flight/apps/adcs/sun.py
```diff
@@ -82,21 +80,28 @@ def compute_body_sun_vector_from_lux(I_vec):
...
+    I_valid = [I_vec[idx] for idx in valid_sensor_idxs]
+    print(N_valid)
+
+    # Compute the Inverse of the valid light sensor normals using the Moore-Penrose pseudo-inverse
+    oprod_sun_inv = invert_3x3_psd(np.dot(N_valid.transpose(), N_valid))
+    if oprod_sun_inv is None:  # If the inverse is not possible, sun positioning is not uniquely determinable
+        status = StatusConst.SUN_NOT_ENOUGH_READINGS
+        return status, sun_body
+    else:
+        oprod_sun_inv = np.dot(oprod_sun_inv, N_valid.transpose())
```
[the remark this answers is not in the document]
kmkar31: Pre-allocated all matrices

flight/apps/adcs/sun.py
```diff
@@ -82,21 +80,28 @@ def compute_body_sun_vector_from_lux(I_vec):
...
-    sun_body[0] = i_vec[0] if i_vec[0] > i_vec[1] else -i_vec[1]
-    sun_body[1] = i_vec[2] if i_vec[2] > i_vec[3] else -i_vec[3]
-    sun_body[2] = i_vec[4]
+    # Extract body vectors and lux readings where the sensor readings are valid
+    valid_sensor_idxs = [
+        idx for idx in range(len(I_vec)) if I_vec[idx] != ERROR_LUX and I_vec[idx] >= THRESHOLD_ILLUMINATION_LUX
+    ]
+    N_valid = PhysicalConst.LIGHT_SENSOR_NORMALS[valid_sensor_idxs, :]
+    I_valid = [I_vec[idx] for idx in valid_sensor_idxs]
+    print(N_valid)
```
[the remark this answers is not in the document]
kmkar31: Done

flight/apps/adcs/sun.py
```diff
@@ -72,42 +74,38 @@ def compute_body_sun_vector_from_lux(I_vec):
...
-    sun_body[0] = i_vec[0] if i_vec[0] > i_vec[1] else -i_vec[1]
-    sun_body[1] = i_vec[2] if i_vec[2] > i_vec[3] else -i_vec[3]
-    sun_body[2] = i_vec[4]
-
-    norm = (sun_body[0] ** 2 + sun_body[1] ** 2 + sun_body[2] ** 2) ** 0.5
-    # norm = MAX_RANGE
+    # Extract body vectors and lux readings where the sensor readings are valid
+    ACTIVE_LIGHT_SENSORS = np.array(I_vec) > _THRESHOLD_ILLUMINATION_LUX
+    ACTIVE_LIGHT_NORMALS = np.array(
+        [PhysicalConst.LIGHT_SENSOR_NORMALS[i] for i in range(_NUM_LIGHT_SENSORS) if ACTIVE_LIGHT_SENSORS[i]]
```
kmkar31, line 93: I absolutely **hate** doing list comprehension for this, but CircuitPython does not let me slice arrays using other arrays

Ibrassow: Preallocate and hardcode the whole array with conditions if necessary. Dynamic allocations are an evil to be eradicated.

kmkar31: Wait, this is not a dynamic allocation. ACTIVE_LIGHT_NORMALS and ACTIVE_LIGHT_SENSORS are variables that are pre-defined at the top of this file. They may change in size (i.e., reduce in memory consumption), but not increase

Ibrassow: Dynamic allocation means not static because it reallocates the memory (free then malloc again). It's not reusing the same buffer. Even if smaller so you incur the cost of reallocating on the heap (unless there is some python magic which I doubt)

kmkar31: I have no idea how to avoid this though. We'd need to hardcode $$9!$$ combinations of light sensor normals to be able to not dynamically allocate. At that point, the memory consumption will be so large that it won't matter

kmkar31:
> Was it tested on HW?

Tested on MBv3 and works

flight/apps/adcs/sun.py
kmkar31: Tried If logic. It is way too complicated will introduce more bugs. Working code more important at this point compared to memory efficiency

Ibrassow (dismissed):
Fix linting

flight/apps/adcs/sun.py
```diff
@@ -72,42 +69,44 @@ def compute_body_sun_vector_from_lux(I_vec):
...
-    sun_body = sun_body / norm
-
-    return status, sun_body
+    # Extract body vectors and lux readings where the sensor readings are valid
+    ACTIVE_LIGHT_READINGS = [I_vec[i] for i in range(_NUM_LIGHT_SENSORS) if I_vec[i] > _THRESHOLD_ILLUMINATION_LUX]
+    ACTIVE_LIGHT_NORMALS = np.array(
+        [PhysicalConst.LIGHT_SENSOR_NORMALS[i] for i in range(_NUM_LIGHT_SENSORS) if I_vec[i] > _THRESHOLD_ILLUMINATION_LUX]
+    )
+
+    try:
```
Akshat412, line 92: If the try-except cannot be an if-else, add a comment explaining why.

Akshat412 (changes requested):

Ibrassow (dismissed):

perrintong325 (dismissed):

Akshat412 (dismissed):

Ibrassow (approved):

Akshat412 (approved):

perrintong325 (approved):

PR #157, opened 2025-02-16
Move thermal to EPS + EPS bug fixes

achstar:
Getting rid of thermal task and moving all temp logging to EPS.


achstar:
untested. Haven't updated packing.py yet, and thermal.py is not yet deleted

flight/hal/drivers/adm1176.py
```diff
@@ -56,7 +56,7 @@ class ADM1176:
     def __init__(self, i2c_bus, addr=0x4A):
         self.i2c_device = I2CDevice(i2c_bus, addr, probe=False)
         self.i2c_addr = addr
-        self.sense_resistor = 1
+        self.sense_resistor = 0.01
```
achstar, line 59: Might want to double check that this is 0.01 ohms, the board current and radio current values are looking a bit high imo. (~20 mA for radio when I'm fairly sure it's not on, ~50-200+ mA for board current which could be accurate? Need another opinion on that.)

flight/tasks/eps.py
```diff
@@ -84,6 +87,7 @@ def read_fuel_gauge(self):
             self.log_data[EPS_IDX.BATTERY_PACK_MIDPOINT_VOLTAGE] = int(fuel_gauge.read_midvoltage())
             self.log_data[EPS_IDX.BATTERY_PACK_TTE] = int(fuel_gauge.read_tte())
             self.log_data[EPS_IDX.BATTERY_PACK_TTF] = int(fuel_gauge.read_ttf())
+            self.log_data[EPS_IDX.BATTERY_PACK_TEMPERATURE] = int(fuel_gauge.read_temperature() * 100)
```
achstar, line 90: I thought it might be good to have this rounded to the nearest hundredth Celsius but also open to just having it be the nearest Celsius

flight/tasks/eps.py
```diff
@@ -95,7 +99,7 @@ async def main_task(self):
         else:
             if not DH.data_process_exists("eps"):
                 data_format = (
-                    "Lhhb" + "h" * 4 + "L" * 2 + "h" * 30 + "b"
+                    "Lbhhhhb" + "h" * 4 + "L" * 2 + "h" * 30
```
achstar, line 102: will update the spreadsheet once this PR is close to merge

achstar:
Tested on hardware, everything looks good. Temperature readings for fuel gauge are within a few degrees Celsius of CPU temp readings which makes sense

flight/hal/drivers/adm1176.py
Akshat412: I think 20 mA of current for the radio is reasonable; it's always in RX mode, unless it is actively transmitting when it is in TX mode (where the spikes of 500 mA occur).

The board power includes every 3.3V device on the mainboard so that is also probably reasonable? Still good to check those resistor values though.

flight/tasks/eps.py
```diff
@@ -123,14 +127,18 @@ async def main_task(self):
                         + f"Radio Current: {self.log_data[EPS_IDX.RF_LDO_OUTPUT_CURRENT]} mA"
                     )
 
+            self.log_data[EPS_IDX.CPU_TEMPERATURE] = int(microcontroller.cpu.temperature * 100)
+            self.log_info(f"CPU temperature: {self.log_data[EPS_IDX.CPU_TEMPERATURE] / 100}° ")
+
             if self.read_fuel_gauge():
+                self.log_info(f"Battery Pack Temperature: {self.log_data[EPS_IDX.BATTERY_PACK_TEMPERATURE] / 100}° C")
```
Akshat412, line 134: I think these were being scaled up to 100ths in the old thermal task, but I don't love selective scaling for some data in downlinked messages. If the resolution is important info these can definitely be floats, but if not we should just limit them to integer values.

flight/hal/drivers/max17205.py
```diff
@@ -191,15 +193,28 @@ def read_time_pwrup(self):
...
+        """
+        Reads the temperature of the battery pack.
+
+        :return: Temperature of the battery pack in Celsius
+        """
+        with self.i2c_device as i2c:
+            i2c.write(bytes([_MAX1720X_TEMP_ADDR]))
+            i2c.readinto(self.rx_buffer)
+
+        self.temperature = int.from_bytes(self.rx_buffer, "little", signed=False) / 256.0
```
Ibrassow, line 215: prefer multiplying by 0.00390625 instead of the division. (unrelated: I'd have done bit shift if we didn't care about the decimals)

flight/tasks/eps.py
Ibrassow: I leave the decision up to you

flight/tasks/eps.py
Ibrassow: 1. prefer multiplication over divisions in general (*0.01) 
2. The scaling compresses those values into 2-byte (short) integers instead of 4-byte floats. This reduces file size, saving 2 bytes per log. As a result, we can store more data lines per file for the same file size as you scale up (TM space savings is just another nice result)

Ibrassow: also, no need to divide in the log (just print the result in ° cC (I know it's weird but that makes the logging faster)

Ibrassow (changes requested):
Minor minor changes but after that, consider it approved

flight/hal/drivers/adm1176.py
Akshat412: Actually, given the supply reports about 60 mA when the radio is not transmitting, the spikes to 200 mA are a little concerning. They may very well exist but I'd like to know that. We should get those Nordic Power Profilers in.

flight/tasks/eps.py
Akshat412: I'd go as far as to officially make the logged data centi-Celsius, and also reflect this in the spreadsheet. My only concern was some places thinking of this data as Celsius and some centi-Celsius.

Akshat412 (changes requested):

flight/hal/drivers/adm1176.py
Akshat412: Yeah sorry I was referring to the power supply reading. Given it's usually around 50-60 mA, the spikes of 200 mA are what were concerning, and will not be picked up by the supply's display since it cannot respond fast enough. Could be coming from SD card write operations? Not sure. 

We should try to measure the current draw using an external device while the FSW runs, and compare logs / operations to power consumption. As long as those spikes are repeatable / explainable, it should be fine, but if they're randomly occuring that's kinda weird.

Ibrassow (approved):

PR #161, opened 2025-02-17
ADCS Bug fixes

kmkar31:
First (of many) bug fixes
1. ```attitude_control()``` in adcs.py computes the unbiased gyro reading and is supposed to subtract the bias from the measured gyro reading. However, I subtracted the gyro reading from the gyro reading due to an indexing error. ACS always received a zero angular velocity and did nothing
2. In ```initialize_mekf()```, sun status is checked to see if a valid sun position is received. In the IF condition, I forgot to add ```sun_status == SUN_VECTOR_STATUS.ECLIPSE``` and instead had ```if SUN_VECTOR_STATUS.ECLIPSE``` which was always true and resulted in MEKF initialization always failing
3. Remnant of the old logging structure in ```EKF_update()```. Switched the logging framework to the new one

Reason why these bugs were not found:
- Only Bug 3 will actually throw an error. Bug 1 and 2 will only do incorrect things. However, because of Bug 2, MEKF never initializes and therefore Bug 3 never shows up

### DIFF
--- a/flight/apps/adcs/ad.py
+++ b/flight/apps/adcs/ad.py
@@ -19,7 +19,6 @@ from apps.adcs.orbit_propagation import OrbitPropagator
 from apps.adcs.sun import SUN_VECTOR_STATUS, approx_sun_position_ECI, compute_body_sun_vector_from_lux, read_light_sensors
 from apps.telemetry.constants import GPS_IDX
 from core import DataHandler as DH
-from core import logger
 from hal.configuration import SATELLITE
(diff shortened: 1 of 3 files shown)


Ibrassow (approved):
Not fond of all those lists of strings for the error messages... because it's not apparent at first glance it's a logging. Think you should have it in place, despite the redundancies...

Letting @Akshat412 handle the HW tests

Akshat412 (approved):
Nice work!!!!

PR #180, opened 2025-02-23
SIL CI Integration

atletea:
- Added sil.yml file that sets up and runs the simulation for each push on Github workflow
- Added sil_run.py that runs the simulation for 30 seconds 
- Added parse_sillogs.py which will parse logs and stop the workflow from successful execution if there are errors present in the simulation run 
The parse_sillogs.py can be modified depending on what we want to error check for, for now the workflow should work for any commit/push.


parse_sillogs.py
```diff
@@ -0,0 +1,19 @@
...
+            python_err = True
+        if "[ERROR]" in line:
+            sim_err = True
+        if python_err or sim_err:
+            error_log.write(line)
+
+if python_err:
+    raise Exception("Python errors found, see error log.")
+if sim_err:
+    raise Exception("Sim errors found, see error log.")
```
Ibrassow, line 19: - For both python scripts, put everything under a if __name__ == "__name__": ... 
- For the configuration, use command line arguments: time duration, filename, etc so we can generalize those scripts for other purposes.

sil_run.py
```diff
@@ -0,0 +1,19 @@
...
+import time
+
+log_filename = "sil_sim.log"
+# error_log_filename = f"{timestamp}_err.log"
+
+try:
+    with open(log_filename, "w") as log_file:
+        process = subprocess.Popen(["./run.sh", "simulate"], stdout=log_file, stderr=log_file, preexec_fn=os.setsid)
+        print(f"Running simulation for 120 seconds, output written to {log_filename}")
+        time.sleep(30)
```
Ibrassow, line 13: see comment above

parse_sillogs.py
```diff
@@ -0,0 +1,19 @@
...
+sim_err = False
+with open(log_filename, "r") as log_file, open(error_filename, "w") as error_log:
+    for line in log_file:
+        # errors running the SIL
+        if "Traceback" in line:
+            python_err = True
+        if "[ERROR]" in line:
+            sim_err = True
+        if python_err or sim_err:
+            error_log.write(line)
```
Ibrassow, line 14: you can have those keywords defined in a config file (.txt or whatever) that your scripts will basically read to know all triggers (e.g. so people could add "WARNING" to the list easily)

.github/workflows/sil.yml
```diff
@@ -0,0 +1,77 @@
...
+          name: ${{ env.TIMESTAMP }}-sil-logs
+          path: sil_sim.log
+      - name: Parse logs for errors
+        run: python3 parse_sillogs.py
+      - name: Upload error log file
+        uses: actions/upload-artifact@v4
+        if: always()
+        with:
+          name: ${{ env.TIMESTAMP }}-sil-errs
+          path: sil_err.log
```
Ibrassow, line 76: add some spacing between the job names for clarity

Ibrassow (changes requested):

sil_run.py
kmkar31, line 13: At the current frequency (GPS runs at 0.03 Hz) , I'd rather run this script for ~150s or so

kmkar31 (changes requested):
Agree with Ibrahima's comments. Nothing else to add

sil_run.py
Ibrassow: good point 

Honestly, @atletea go for > 5min

kmkar31 (dismissed):
Since my changes are the same as Ibrahima's, I'll approve so I don't block a merge

Akshat412:
My changes are also the same as Ibrahima's. Everything else looks good!

For actually testing FSW with this, we can temporarily add a try catch in every task, essentially some method to raise every error in FSW as an ERROR or WARNING. IDK if there's a clean way of doing this but it should be done periodically to fully utilize this script.

parse_sillogs.py
kmkar31: Fixed

.github/workflows/sil.yml
kmkar31: Done

sil_run.py
kmkar31: Done

kmkar31: Done

Akshat412:
What about individually within each task, inside the task's main function?

kmkar31:
Thanks for the idea. Changed the debug process inside TemplateTask to throw log.error instead of log.info. This should not affect flight ops since debug will be False then.

SIDE NOTE: There is an error inside the Command task which now throws an error and therefore the SIL CI test fails here

parse_sillogs.py
kmkar31: Done. I have a dictionary inside sil_run.py that has different keywords and mapping to colors in the log file. Additioanl keywords can be added there. There is specific handling for 'ERROR' alone that throws an exception and causes the CI test to fail

Ibrassow (approved):
Great

Ibrassow:
> Thanks for the idea. Changed the debug process inside TemplateTask to throw log.error instead of log.info. This should not affect flight ops since debug will be False then.
> 
> SIDE NOTE: There is an error inside the Command task which now throws an error and therefore the SIL CI test fails here

I assume this is fixed?

kmkar31:
Yep, all good now

PR #184, opened 2025-02-25
New HAL structure for boot and reboot sequence

perrintong325:
-New Structure
    - New device class that has 3 components: boot_fn, error and device object
    - All devices stored inside a device list in cubesat class
    - each satellite can append specific devices to this list, (eg Argus2 doesn't have extra features not present on other boards, so no append in Argus2 init. Argus1.1 has neopixel, which is appeneded in init)
    - ** interface remains the same outside of HAL**
    - VFS removed as a separate boot fn, now integrated with SD
    - charger removed completely cos that was just useless
-Error handling
    - reboot fn de-instantiates devices
-Build Script
    - Supports both CPY9 and CPY8, change the CPY version in build.py to build for the targeted platform
 
Tested on HW with 4 XY boards, emulator changes done but untested


flight/tasks/hal_monitor.py
```diff
@@ -0,0 +1,26 @@
...
+    def __init__(self, id):
+        super().__init__(id)
+        self.name = "HAL_MONITOR"
+
+    async def main_task(self):
+
+        if SM.current_state == STATES.STARTUP:
+            for name, device in SATELLITE.ERRORS.items():
+                if device[0] == 1:
+                    SM.set_state(STATES.HAL_ERROR)
```
Akshat412, line 21: IDT we have a HAL_ERROR state right now. We can but it needs to be added in states,py and in the statechart.
Also if we add it we should have a very good, very deterministic set of things we do in that state to recover from HAL errors.

flight/hal/cubesat.py
```diff
@@ -10,59 +10,52 @@ class CubeSat:
...
+            "SDCARD": Device(self.__sd_card_boot),
+            "IMU": Device(self.__imu_boot),
+            "RTC": Device(self.__rtc_boot),
+            # "GPS": Device(self.__gps_boot),
+            # "RADIO": Device(self.__radio_boot),
+            # "FUEL_GAUGE": Device(self.__fuel_gauge_boot),
+            # "BURN_WIRE": Device(self.__burn_wire_boot),
+            "BOARD_PWR": Device(self.__power_monitor_boot),
+            # "RADIO_PWR": Device(self.__power_monitor_boot),
+            # "JETSON_PWR": Device(self.__power_monitor_boot),
```
Akshat412, line 28: GPS / peripheral power monitor is missing.

emulator/cubesat.py
```diff
@@ -25,22 +25,41 @@ def __init__(self):
...
         self._payload_uart = None
+        self._device_list = {
+            "SDCARD": Device(None),
+            "IMU": Device(None),
+            "RTC": Device(None),
+            "GPS": Device(None),
+            "RADIO": Device(None),
+            "FUEL_GAUGE": Device(None),
+            "BURN_WIRE": Device(None),
+            "BOARD_PWR": Device(None),
```
Akshat412, line 38: Similarly, peripheral / GPS power monitor is missing.

Akshat412 (changes requested):

flight/tasks/hal_monitor.py
perrintong325: removed for this pr

perrintong325 (commented):
removed hal monitor for this pr, included gps power monitor

flight/hal/argus_v1_1.py
```diff
@@ -371,122 +330,82 @@ def __imu_boot(self) -> list[int]:
...
+            "LIGHT_XP": [ArgusV1Components.LIGHT_SENSOR_XP_I2C_ADDRESS, ArgusV1Components.LIGHT_SENSOR_X_I2C],
+            "LIGHT_XM": [ArgusV1Components.LIGHT_SENSOR_XM_I2C_ADDRESS, ArgusV1Components.LIGHT_SENSOR_X_I2C],
+            "LIGHT_YP": [ArgusV1Components.LIGHT_SENSOR_YP_I2C_ADDRESS, ArgusV1Components.LIGHT_SENSOR_Y_I2C],
+            "LIGHT_YM": [ArgusV1Components.LIGHT_SENSOR_YM_I2C_ADDRESS, ArgusV1Components.LIGHT_SENSOR_Y_I2C],
+            "LIGHT_ZM": [ArgusV1Components.LIGHT_SENSOR_ZM_I2C_ADDRESS, ArgusV1Components.LIGHT_SENSOR_Z_I2C],
+            "SUN1": [ArgusV1Components.SUN_SENSOR_ZP1_I2C_ADDRESS, ArgusV1Components.SUN_SENSOR_ZP_I2C],
+            "SUN2": [ArgusV1Components.SUN_SENSOR_ZP2_I2C_ADDRESS, ArgusV1Components.SUN_SENSOR_ZP_I2C],
+            "SUN3": [ArgusV1Components.SUN_SENSOR_ZP3_I2C_ADDRESS, ArgusV1Components.SUN_SENSOR_ZP_I2C],
+            "SUN4": [ArgusV1Components.SUN_SENSOR_ZP4_I2C_ADDRESS, ArgusV1Components.SUN_SENSOR_ZP_I2C],
         }
```
Ibrassow, line 386: Use LIGHT_ZP_X. A sun sensor is a totally different sensor. THose are still individual light sensors

flight/hal/argus_v2.py
```diff
@@ -330,234 +274,155 @@ def __gps_boot(self) -> list[int]:
...
                 raise e
+            return [None, Errors.ADM1176_NOT_INITIALIZED]
 
-            return Errors.IMU_NOT_INITIALIZED
-
-        return Errors.NOERROR
-
-    def __charger_boot(self) -> list[int]:
-        """charger_boot: Boot sequence for the charger
+    def __imu_boot(self, _) -> list[object, int]:
```
Ibrassow, line 334: if you adapt the argument list like this, you need to do it for all devices (e.g. burn wire, rtc, etc)

flight/hal/argus_v2.py
```diff
@@ -707,11 +543,21 @@ def __fuel_gauge_boot(self) -> list[int]:
...
+
+        :param peripheral: The peripheral to reboot
+        :return: Error code if the reboot failed
+        """
+        if device_name not in self.__device_list:
+            return Errors.PERIPHERAL_NOT_FOUND
+        self.__device_list[device_name].device = None
+        func = self.__device_list[device_name].boot_fn
+        self.__device_list[device_name].device, self.__device_list[device_name].error = func(device_name)
+        return self.__device_list[device_name].error
```
Ibrassow, line 563: 1. Has it been tested extensively on the board with an actual failure? For example, disconnecting one of the X/Y boards to see if the error logic correctly "disables" that device. Can you confirm that it successfully reboots and returns to nominal behavior after reconnecting?

2. Setting it to None does not free the memory because Python uses reference counting. If any variable in the codebase still references the object, it won't be deallocated (e.g., the radio object used in the comms apps, @Akshat412 ). We should avoid this kind of referencing throughout the codebase and enforce the use of SATELLITE.xx instead. Would make the changes in this PR

3. Have you checked the deinit() method available in CircuitPython? Similar to point 2, your pins might remain active, and a proper reboot would release and re-enable them. Typically, a de-init/destructor procedure ensures graceful de-initialization. I suggest implementing a deinit() or post_init() function for each driver (abstract in Device) to properly release the pins per driver. This might be useful: https://docs.circuitpython.org/en/latest/shared-bindings/busio/ 

pinging others that are concerned @kmkar31 @achstar

Ibrassow: btw, most of time, the driver itself (if not, datasheet) has either an implementation or indications at least

flight/hal/drivers/drv8235.py
```diff
@@ -0,0 +1,325 @@
...
+    """
+
+
+#     @property
+#     def get_flags(self):
+#         flags = {}
+#         if self._fault:
+#             if self._stall:
+#                 flags["stall"] = None
+#             if self._ocp:
```
Ibrassow, line 256: if we don't use, remove those commented blocks

Ibrassow (changes requested):
I cleaned up a bit the git history. Please, keep the commit msg focused on the particular feature change as clearly as possible. 

Also, I'd avoid having a long dev branch in which you keep merging, this makes PRs misleading and confusing. Also, we should prefer rebasing branches rather than merge with main

flight/hal/argus_v2.py
perrintong325: can you clarify what you are talking about? all of them are identical

flight/hal/drivers/drv8235.py
perrintong325: this driver is a placeholder. it works but im going to go through it line by line again cos freeda claimed it was tested but i know there's no way it was actually tested cos it instantiates i2c twice ie it would never run.

emulator/cubesat.py
```diff
@@ -25,22 +25,41 @@ def __init__(self):
...
-        self._torque_drivers = {}
-        self._light_sensors = {}
-        self._fuel_gauge = None
-        self._rtc = None
-        self._radio = None
-        self._sd_card = None
-        self._burn_wires = None
-        self._vfs = None
         self._payload_uart = None
+        self._device_list = {
```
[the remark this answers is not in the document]
perrintong325: no gain or loss to dict itself. overall gained 1% because we used to store each object twice for no reason, we also know store individual errors so that probably would hv consumed some ram. but yes overall gained 1%

flight/hal/cubesat.py
```diff
@@ -183,160 +161,157 @@ def IMU_TEMPERATURE_AVAILABLE(self) -> bool:
...
         :return: object or None
         """
         return self._time_ref_boot
+
+    def reboot_peripheral(self, peripheral: object) -> int:
+        """REBOOT_PERIPHERAL: Reboot the peripheral"""
+        raise NotImplementedError("CubeSats must implement reboot peripheral method")
+
+
+class Device:
```
Akshat412, line 313: Class device is not near entities that use it, which reduces clarity.

flight/hal/cubesat.py
```diff
@@ -10,59 +10,53 @@ class CubeSat:
...
+        self.__device_list = {
+            "SDCARD": Device(self.__sd_card_boot),
+            "IMU": Device(self.__imu_boot),
+            "RTC": Device(self.__rtc_boot),
+            # "GPS": Device(self.__gps_boot),
+            # "RADIO": Device(self.__radio_boot),
+            # "FUEL_GAUGE": Device(self.__fuel_gauge_boot),
+            # "BURN_WIRE": Device(self.__burn_wire_boot),
+            "BOARD_PWR": Device(self.__power_monitor_boot),
+            # "RADIO_PWR": Device(self.__power_monitor_boot),\
```
Akshat412, line 27: Backslash

flight/hal/argus_v2.py
```diff
@@ -276,52 +252,20 @@ def __init__(self, debug: bool = False):
...
-            print()
-            for error in error_list:
-                print(f"{Errors.diagnostic_to_string(error)}")
-            print()
-
-        self.__recent_errors = error_list
-
-        return error_list
+
+        for name, device in self.__device_list.items():
```
Akshat412, line 258: Check the order of devices booting please

flight/hal/argus_v2.py
Akshat412: This should be in a different PR. HAL monitor and error handling is a whole another can of worms.


Also I think deinit just deinitializes GPIO pin assignments so they can be reused. We still need to find an appropriate way of deallocating memory.

Akshat412 (changes requested):

flight/hal/drivers/errors.py
```diff
@@ -0,0 +1,24 @@
+from micropython import const
+
+
+class Errors:
```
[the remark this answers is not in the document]
perrintong325: these will be public, taking it out it's only used for hal monitor

flight/hal/argus_v2.py
Ibrassow: No, the deinit must be in this PR. We,re talking about simply deinitializing pins in a deinit fctn per driver.

flight/hal/argus_v2.py
Ibrassow: My bad, I probably checked a different diff. __state_flags_boot is still only with self (missing the second argument "_")

flight/hal/argus_v2.py
Akshat412: Oh, yeah deinit-ing hardware should be in this PR. I will make the changes to the comms application to remove cls.sat in #186 to support this method of deallocation.

Ibrassow: Would suggest you make the change in this PR to keep commits localized and related

Ibrassow:
note: @perrintong325 update PR name and description to align with what is actually being dev here

scripts/li ght_sensor_data_collection
```diff
@@ -0,0 +1,70 @@
+import time
```
Akshat412, line 1: Delete this script or fix it, please

Akshat412 (approved):
Nice work

perrintong325:
small change to imu noise test script to print boot errors correctly

flight/hal/drivers/adm1176.py
```diff
@@ -70,6 +73,16 @@ def __init__(self, i2c_bus, addr=0x4A):
...
+    def reset(self) -> None:
+        """reset: Resets the device and clears all registers.
+
+        :return: None
+        """
+        self.__turn_off()
+        time.sleep(0.1)
+        self.__turn_on()
+        time.sleep(0.1)
+
```
Ibrassow, line 85: Commented on this in #192, thus I expect it to be fixed in that one.

Ibrassow (approved):

Ibrassow (approved):

Akshat412 (approved):

PR #186, opened 2025-02-26
CDH Integration for all Data Downlinks and Command Uplinks

Akshat412:
Complete integration between command processor and comms for execution of uplinked commands. Command processor now also controls the payload for downlinked data.

Specific additions:
- For commanding a specific event on the SC (reboots, time reference updates, etc.), the SC downlinks an ACK indicating whether or not the command was successfully executed.
- For TM frame requests, the command processor packs the requested telemetry frame which is then downlinked.
- For file transfers, the command processor sets the comms filepath definition to the requested file. Comms handles the transport protocol to downlink the file internally.
- TM frame definition for TM_STORAGE, with DH updates to return storage info per task.


flight/apps/command/commands.py
```diff
@@ -42,16 +52,22 @@ def SWITCH_TO_STATE(target_state_id, time_in_state=None):
...
 
 
-def UPLINK_ORBIT_REFERENCE(time_in_state, orbital_parameters):
+def UPLINK_ORBIT_REFERENCE(time_reference, orbital_parameters):
     """Sends time-referenced orbital information to update the orbit reference."""
     logger.info(
-        f"Executing UPLINK_ORBIT_REFERENCE with orbital_parameters: {orbital_parameters}, time_in_state: {time_in_state}"
+        f"Executing UPLINK_ORBIT_REFERENCE with orbital_parameters: pos({orbital_parameters}, time_reference: {time_reference}"
+    )
+    OrbitPropagator.propagate_orbit(
```
Akshat412, line 69: @kmkar31 you'd mentioned this should be handled differently. Does the orbit propagator have a mechanism for receiving this data?

kmkar31: You don't need to do this. You can directly update ```last_update_time``` and ```last_updated_state``` attributes of the class. You can be pedantic and add a class method inside OrbitProp to do the same, but I leave that to you.

Operational details I need to know about: 

1. When you uplink a position reference, how long ago would that position be compared to the satellite's current position? If it is planned to be fairly old (like > few minutes), then we may have to reconsider how we use uplinks

2. When the satellite restarts, are the old uplinked commands just deleted or are they stored? Put another way, we need to re-instantiate the class to its last computed state & time to not lose progress and do something non-sensical

3. I need to have orbit checks to make sure we don't use the wrong units. Do we want them on the ground or onboard

kmkar31 (changes requested):
Have reviewed changes to adcs / adcs related files and nothing else

flight/apps/command/commands.py
Akshat412: 1. This would probably depend on how the orbit reference is calculated on the ground in the first place which I don't know much about. In terms of latency between the reference being calculated on the ground and getting to the orbit prop class, given comms latency and task priorities for comms, command and ADCS, I don't see this being more than a minute or two at worst.
2. I think CDH stores command logs as a task called cmd_logs so technically the command is recoverable.
3. Both. We will ensure this stuff is checked before uplinking but best to have a check in FSW too.

Akshat412: Also, we might just want to store the orbit parameters currently in use through the ADCS task logs, right?

kmkar31: Interesting, I hadn't thought about reading from the adcs log for it. I can probably add that functionality into orbit prop

Akshat412: Let's do that separately, keeping this comms / command focused is better

kmkar31 (dismissed):
Will handle OrbitProp type checking and re-initialization in another PR

Ibrassow:
> Will handle OrbitProp type checking and re-initialization in another PR

reference the PR here

tests/test_data_handler.py
```diff
@@ -68,6 +68,27 @@ def test_extract_time_from_filename(input_filename, expected_output):
...
+    files = [
+        "imu_1700001234.bin",
+        "cdh_1739719847.bin",
+        "thermal_1739720103.bin",
+        "img_1699999999.bin",
+        "comms_1589387566.bin",
+    ]
+    assert get_closest_file_time(file_time, files) == expected_output
+
+
```
Ibrassow, line 91: need to test cases where it fails too (e.g. the time isn't in the filename)

Ibrassow (changes requested):

flight/apps/telemetry/packing.py
```diff
@@ -336,13 +341,175 @@ def pack_tm_heartbeat(cls):
...
+
+            else:
+                logger.warning("No latest CDH data available")
+
+        else:
+            logger.warning("No CDH data available")
         pass
 
     @classmethod
     def pack_tm_storage(cls):
```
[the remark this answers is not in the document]
Akshat412: AFAIK this can be decoded and then uploaded to the database correctly.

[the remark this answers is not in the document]
Ibrassow: Do you get the right values though? (can print it on the spacecraft to compare)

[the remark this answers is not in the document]
Akshat412: This I'm unsure of. Let me test and get back to you.

perrintong325 (dismissed):

flight/apps/command/preconditions.py
```diff
@@ -0,0 +1,53 @@
...
+    # Check that time is not in the future
+    if time_reference > time.time():
+        return False
+
+    # Get the timestamp for 7 days ago
+    past_week = time.time() - (_SECONDS_IN_WEEK)
+
+    # Check that time is not from before the past week
+    if time_reference < past_week:
+        return False
```
Ibrassow, line 38: I don't think you want to have those conditions like that. If the SC gets its time out of sync (e.g. RTC failure), your time will always be in the future. Even with normal behavior, the time can be in the future. You want to test if the time ref is in the correct UNIX format + we're not asking for time before launch or before 2025 for instance. 

For any preconditions, be wary of any conditions that prevent your control of the SC. Also, we should never assume your SC is in a specific state (like having roughly its time coherent)

cc @kmkar31 
@Akshat412

Ibrassow (approved):
Good for merge!

Issue #172, opened 2025-02-20
CDH: add large buffer in ImageProcess to improve inter-board data rate


Ibrassow:
Solving this issue will improve the inter-board data rate (MB<->Payload). 

The SD writes by blocks of 512 bytes so we can pre-allocate that static buffer in ImageProcess and put all the data in until it is full for a write. Then, we just need to reset the counting variables/indices and continue overwriting. It's basically a circular buffer with a single write once we reach a full length.

PR #187, opened 2025-03-01
CDH: large buffer in ImageProcess

s026134:
Issue: #172

Added a large buffer to the ImageProcess class for the Onboard Data Handler to improve inter-board data rate by writing to the SD card in blocks of 512 bytes instead of every time we receive data. It is a circular buffer that will collect the logs and then when it hits 512 bytes in the buffer it will write to a file. Any remaining data will be added to the buffer after the write and index variable keeps track of the current size.

### DIFF
--- a/flight/core/data_handler.py
+++ b/flight/core/data_handler.py
@@ -570,7 +570,9 @@ class ImageProcess(DataProcess):
         self.current_path = self.create_new_path()
         self.delete_paths = []  # Paths that are flagged for deletion
         self.excluded_paths = []  # Paths that are currently being transmitted
-        self.circular_buffer_size = 20  # Default size of the circular buffer for the files in the directory
+        self.circular_buffer_size = 512  # Default size of the circular buffer for the files in the directory
+        self.img_buf = bytearray(self.circular_buffer_size)  # Pre-allocated static buffer for image process
+        self.img_buf_index = 0  # index to track position in buffer
 
         config_file_path = join_path(self.dir_path, _PROCESS_CONFIG_FILENAME)
         if not path_exist(config_file_path):
@@ -604,7 +606,7 @@ class ImageProcess(DataProcess):
 
     def log(self, data: bytearray) -> None:
         """
-        Logs the given image data.
+        Logs the given image data. Stores into a buffer and write in blocks of 512 Bytes
 
         Args:
             data (List[bytes]): The bytes of image data to be logged.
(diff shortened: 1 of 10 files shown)


flight/core/data_handler.py
```diff
@@ -570,7 +570,9 @@ def __init__(self, tag_name: str):
         self.current_path = self.create_new_path()
         self.delete_paths = []  # Paths that are flagged for deletion
         self.excluded_paths = []  # Paths that are currently being transmitted
-        self.circular_buffer_size = 20  # Default size of the circular buffer for the files in the directory
+        self.circular_buffer_size = 512  # Default size of the circular buffer for the files in the directory
+        self.img_buf = bytearray(self.circular_buffer_size)  # Pre-allocated static buffer for image process
+        self.img_buf_index = 0  # index to track position in buffer
```
Ibrassow, line 576: use another variable, the circular_buffer_size here is used for the cleaning of the files

flight/core/data_handler.py
```diff
@@ -604,19 +606,39 @@ def create_new_path(self) -> str:
...
+            self.file.flush()
+
+            # Store the over flow data for next time buf is full
+            overflow_len = data_len - space_remaining
+            self.img_buf[:overflow_len] = data[space_remaining:]
+            self.img_buf_index = overflow_len  # set the index to the overflow value
+        else:
+            # Nominal - just adding to the buffer until full
+            self.img_buf[self.img_buf_index : self.img_buf_index + data_len] = data
+            self.img_buf_index += data_len  # increment index for the buffer
```
Ibrassow, line 639: Make all this conditioned on cls.SD_ERROR_FLAG being false, so in the case the SD card fails, we will be able to transmit slowly (and painfully) through the RAM to the ground

tests/test_data_handler.py
```diff
@@ -70,13 +72,37 @@ def test_extract_time_from_filename(input_filename, expected_output):
...
+@pytest.fixture
 def sd_root(tmpdir):
-    sd_root = tmpdir.mkdir("sd")
+    sd_root = tmpdir.mkdir(dh._HOME_PATH)
     return sd_root
 
-def test_scan_sd_card(sd_root):
-    DH.sd_path = str(sd_root)"""
+
+def test_image_log(sd_root):
```
Ibrassow, line 82: Need also to test failing and edge cases

Ibrassow (changes requested):

Ibrassow:
Can you change the target branch to main?

Ibrassow (approved):

PR #196, opened 2025-03-11
ADCS Bug Fixes 3

kmkar31:
PR addresses the following:
1. Adds traceback capability to SIL -> Error traceback is printed when an error is encountered
2. Fixes a bug in GPS time access which led to time being returned as a list
3. Improves ADCS failure logging


Akshat412:
Has this been tested on hardware?

kmkar31:
Will do. however, none of this will throw errors on hardware unless we can get GPS fixes

flight/apps/adcs/consts.py
```diff
@@ -14,36 +14,57 @@ class StatusConst:
...
-    EKF_UPDATE_FAIL = -10
+    MEKF_INIT_FAIL = -1
+    OPROP_INIT_FAIL = -2
+    TRIAD_FAIL = -3
+    POS_UPDATE_FAIL = -4
+    SUN_UPDATE_FAIL = -5
+    MAG_UPDATE_FAIL = -6
+    EKF_UPDATE_FAIL = -7
+    TRUE_SUN_MAP_FAIL = -8
+    TRUE_MAG_MAP_FAIL = -9
```
Ibrassow, line 30: Would be nice to get those codes in some TM frames

Ibrassow (dismissed):

Ibrassow:
Pending hardware testing but lgtm

flight/apps/adcs/consts.py
Akshat412: Agreed, can make this happen. Will make an issue for this.

Akshat412: At the very least, these should be logged to the DH first. Is that done already?

kmkar31: No, multiple of these error codes can occur in a single task execution. How do you want to handle a varying log file?

kmkar31: However, the most important failures like MEKF INIT FAIL, POS, SUN and MAG updates can be added as fields to telemetry and you either get an OK or a fail

Akshat412: Ahh I see, depends on how you'd categorize important / non-important errors. We should figure out how many important errors there are, make an array that is the max size of this, and then always log that array. That way we can easily generate TM / a file from this data.

Akshat412 (dismissed):
Lokks good to me

Akshat412 (approved):
All changes look good

PR #198, opened 2025-03-12
ADCS: Orbit Prop rewrite

kmkar31:
1. Checks for a pre-existing GPS fix from DH to resume orbit prop without having to wait for a GPS fix
2. Re-writes orbit prop to not use Euler integration and instead assume circular orbit. Performs a rotation around omega and is more robust to long periods between GPS fixes
    - Has an associated Unit test for it

PR #205, opened 2025-03-13
Forbid cubesat base class instantiation

Ibrassow:
It should never happen in practice but good to make sure explicitly so it caught in CI.


Ibrassow:
@kmkar31 any idea what happened? This is linked to the submodule hash I assume

kmkar31:
> @kmkar31 any idea what happened? This is linked to the submodule hash I assume

Yep, the submodule was changed to a non-existent commit. Reverted that change. All good now

kmkar31:
Ok, somehow you caught another bug that does not show up in CI. If this is a priority, I can fix that bug in this PR. Else, it is fixed in #198 and will be merged with that

Akshat412 (approved):

perrintong325 (approved):

PR #211, opened 2025-03-13
Fixes to HAL and TM Packer and CDH

Akshat412:
Some bugs that have been caught over the last few days in testing with the GS:

- Changes to TM packing for HBs and TM STORAGE
- Comms now sets internal file ID definition based on filename
- Precondition for UPLINK TIME REFERENCE updated to use existing CPy packages, and have more mission-specific scope
- DH.get_storage_info has more error checking
- CMD Processor test for time preconditions updated to match the new precondition check

### DIFF
--- a/flight/apps/command/preconditions.py
+++ b/flight/apps/command/preconditions.py
@@ -1,5 +1,3 @@
-import time
-
 from apps.command.constants import file_tags_str
 from core.data_handler import DataHandler
 from core.state_machine import STATES
@@ -23,19 +21,15 @@ def valid_time_format(*args) -> bool:
     """
     time_reference = args[0]
 
-    if time_reference < 3004213696:  # 40 years into the future
+    if not isinstance(time_reference, int):
+        return False
+
+    # Check that the timestamp is above Jan 1st 2020
+    # Check that the timestamp is below Jan 1st 2080
+    if 1577854800 < time_reference < 3471310800:
         return True
-    return False
-
-    # if time_reference is None:
-    #     return False
-
-    # try:
-    #     # Try to convert it to a time struct will verify if it was of proper UNIX format
-    #     time.gmtime(time_reference)
-    #     return True
-    # except (ValueError, TypeError, OverflowError, OSError):
-    #     return False
+    else:
+        return False
 
 
 def file_id_exists(*args) -> bool:
--- a/flight/apps/command/preconditions.py
+++ b/flight/apps/command/preconditions.py
@@ -33,9 +33,9 @@ def valid_time_format(*args) -> bool:
     while in development.
     """
 
-    # Check that the timestamp is above Jan 1st 2020
-    # Check that the timestamp is below Jan 1st 2080
-    if 1577854800 < time_reference < 3471310800:
+    # Check that the timestamp is above Jan 1st 2020 UTC
+    # Check that the timestamp is below Jan 1st 2030 UTC
+    if 1577836800 < time_reference < 1893456000:
         return True
     else:
         return False
--- a/flight/apps/command/preconditions.py
+++ b/flight/apps/command/preconditions.py
@@ -23,15 +23,19 @@ def valid_time_format(*args) -> bool:
     """
     time_reference = args[0]
 
-    if time_reference is None:
-        return False
-
-    try:
-        # Try to convert it to a time struct will verify if it was of proper UNIX format
-        time.gmtime(time_reference)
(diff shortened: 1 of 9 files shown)


tests/test_cmd_processor.py
```diff
@@ -221,7 +221,7 @@ def test_valid_time_format():
     assert valid_time_format(1589387500)
 
     # Test edge cases and failing
-    assert valid_time_format(0)
+    assert not valid_time_format(0)  # outside of valid mission range
```
Akshat412, line 224: This is not supported by struct_time in CircuitPython.

flight/apps/command/preconditions.py
```diff
@@ -23,14 +21,23 @@ def valid_time_format(*args) -> bool:
...
+    limit the time reference to at least this date.
+
+    Can also update this to be our launch date,
+    although that will lead to issues in testing
+    while in development.
+    """
+
+    # Check that the timestamp is above Jan 1st 2020
+    # Check that the timestamp is below Jan 1st 2080
+    if 1577854800 < time_reference < 3471310800:
```
perrintong325, line 38: can we double check these values? I'm not confident that I fully know how epoch time actually works in CPY, but these values are only 12AM in EST, they are 5AM in GMT which I'm assuming that's how it is interpreted by default?

flight/apps/telemetry/packing.py
```diff
@@ -81,10 +82,10 @@ def pack_tm_heartbeat(cls):
         if not cls._TM_AVAILABLE:
             cls._TM_AVAILABLE = True
 
-        cls._FRAME = bytearray(227 + 4)  # pre-allocated buffer for packing
+        cls._FRAME = bytearray(229 + 4)  # pre-allocated buffer for packing
```
perrintong325, line 85: magic number

flight/apps/telemetry/packing.py
```diff
@@ -81,10 +82,10 @@ def pack_tm_heartbeat(cls):
...
         cls._FRAME[0] = const(0x01) & 0xFF  # message ID
         cls._FRAME[1:3] = pack_unsigned_short_int([const(0x00)], 0)  # sequence count
-        cls._FRAME[3] = const(227) & 0xFF  # packet length
+        cls._FRAME[3] = const(229) & 0xFF  # packet length
```
perrintong325, line 88: magic number

flight/apps/telemetry/packing.py
```diff
@@ -346,10 +347,10 @@ def pack_tm_hal(cls):
             cls._TM_AVAILABLE = True
 
         # TODO: Frame definition for TM_HAL
-        cls._FRAME = bytearray(13 + 4)  # pre-allocated buffer for packing
+        cls._FRAME = bytearray(14 + 4)  # pre-allocated buffer for packing
```
perrintong325, line 350: magic number

flight/apps/telemetry/packing.py
```diff
@@ -346,10 +347,10 @@ def pack_tm_hal(cls):
...
         cls._FRAME[0] = const(0x02) & 0xFF  # message ID
```
perrintong325, line 351: magic number

perrintong325 (changes requested):

flight/apps/command/preconditions.py
Akshat412: Good catch. We're sticking with UTC time for the mission, so I'll update these.

Akshat412:
This is a good opportunity to start removing magic numbers where possible lol. Good catch on those, I will add constants for all message lengths.

Akshat412:
Fixed all magic numbers.

s026134 (changes requested):

flight/apps/command/preconditions.py
```diff
@@ -23,14 +21,23 @@ def valid_time_format(*args) -> bool:
...
+    limit the time reference to at least this date.
+
+    Can also update this to be our launch date,
+    although that will lead to issues in testing
+    while in development.
+    """
+
+    # Check that the timestamp is above Jan 1st 2020 UTC
+    # Check that the timestamp is below Jan 1st 2030 UTC
+    if 1577836800 < time_reference < 1893456000:
```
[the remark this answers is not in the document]
Akshat412: Yeah fair enough

s026134 (approved):

perrintong325 (approved):

PR #214, opened 2025-03-15
GPS <-> ADCS interface

kmkar31:
1. Adds ECI fields to the GPS logging. These fields are not added to telemetry and do not modify ```packing.py```. By adding ECI to DH handling, we save ADCS from doing a ECEF2ECI conversion at every task execution (~1Hz)
2. Adds an ECEF to ECI state conversion function. This accounts for Earth angular velocity when computing ECI velocity
3. Ensures SIL logs are uploaded even if SIL errors out


Ibrassow (commented):
Can't we save computations by checking if the last orbit reference time has changed? If it hasn't, we just continue with our current state and avoid all the repropagation. Then we don't need to add more confusion with the addition of the ECI state that are not logged

kmkar31:
Ok, I found a better place to put this than GPS task. Will take a while to fix everything though

flight/apps/adcs/frames.py
```diff
@@ -118,3 +119,15 @@ def convert_ecef_to_geoc(ecef, degrees=False):
...
+
+
+def convert_ecef_state_to_eci(r: np.ndarray, v: np.ndarray, utime: int):
+    R_ecef2eci = ecef_to_eci(utime)  # Transformation matrix from ECEF to ECI
+
+    # Transform Position
+    r_eci = np.dot(R_ecef2eci, r)
+
+    # Transform Velocity
+    v_eci = np.dot(R_ecef2eci, v) + np.cross(np.array([0, 0, OMEGA_E]), v)
```
kmkar31, line 131: To be tested with GPS on board. According to this, Earth surface should read a velocity. But we don't. GPS might compensate for surface velocity

kmkar31:
PR ready for review again

flight/apps/adcs/ad.py
```diff
@@ -109,16 +108,14 @@ def read_magnetometer(self) -> tuple[int, int, np.ndarray]:
         """
 
         if SATELLITE.IMU_AVAILABLE:
-            mag = np.array(SATELLITE.IMU.mag())
+            mag = 1e-6 * np.array(SATELLITE.IMU.mag())
```
Akshat412, line 111: Mildly magic number, can you add a comment explicitly stating the conversion happening here?

flight/core/task_configuration.py
```diff
@@ -15,6 +15,6 @@
     TASK.OBDH: {"Task": obdh, "Frequency": 0.5, "Priority": 2},
     TASK.COMMS: {"Task": comms, "Frequency": 1, "Priority": 2, "ScheduleLater": True},
     TASK.ADCS: {"Task": adcs, "Frequency": 5, "Priority": 1},
-    TASK.GPS: {"Task": gps, "Frequency": 0.025, "Priority": 3, "ScheduleLater": True},
+    TASK.GPS: {"Task": gps, "Frequency": 0.25, "Priority": 3, "ScheduleLater": True},
```
Akshat412, line 18: Update this to 0.025 before merge.

Akshat412 (changes requested):

flight/apps/adcs/ad.py
kmkar31: Done

flight/core/task_configuration.py
kmkar31: Done

Akshat412 (dismissed):

Akshat412 (approved):
I think all of my concerns are addressed now. Thanks.

Ibrassow (approved):

PR #217, opened 2025-03-17
Fuel gauge capacity reading change

achstar:
Fuel gauge capacity register was being processed incorrectly. Should just multiply raw register value by 0.5, and that value should fit in 2 bytes


achstar:
not yet hardware tested

Akshat412 (approved):
Looks good!

PR #218, opened 2025-03-17
Add warnings for devices surpassing power consumption thresholds

achstar:
- separate data process for EPS warnings (to be handled by HAL)
- use MAV to calculate avg power consumption for the past second, and use that to compare against power thresholds
- log raw current/voltage values to EPS data process
- time multiplexing: fuel gauge / temp measurements logged at 1 Hz, power monitor / warning readings logged at 5 Hz


achstar:
Needs hw testing

flight/apps/telemetry/packing.py
```diff
@@ -446,6 +455,16 @@ def pack_tm_storage(cls):
...
+        if DH.data_process_exists("eps_warning"):
+            eps_warning_storage_info = DH.get_storage_info("eps_warning")
+            # EPS number of files
+            cls._FRAME[38:42] = pack_unsigned_long_int(eps_warning_storage_info, STORAGE_IDX.NUM_FILES)
+            # EPS directory size
+            cls._FRAME[42:46] = pack_unsigned_long_int(eps_warning_storage_info, STORAGE_IDX.DIR_SIZE)
+        else:
+            logger.warning("EPS Warning Data process does not exist")
+
         ############ ADCS fields ###########
```
achstar, line 468: Oops forgot to adjust indices here, will do that in next push

flight/apps/command/constants.py
```diff
@@ -48,6 +48,7 @@ class CMD_ID:
     8: "thermal",
     9: "gps",
     10: "img",
+    11: "eps_warning"
```
Akshat412, line 51: Good, making this another data process was smart.

flight/apps/telemetry/packing.py
```diff
@@ -42,7 +42,7 @@
 # TM frame sizes as defined in message database
 # TODO: TM HAL
 # TODO: TM PAYLOAD
-_TM_NOMINAL_SIZE = const(227)
+_TM_NOMINAL_SIZE = const(230)
```
Akshat412, line 45: Updating the packer should be in a different PR IMO. Also, these fields may not go in the nominal TM frame, we might want them in the TM HAL frame.

I need to do a PR for implementing TM HAL etc. anyways so I would want to tackle this there.

flight/apps/telemetry/packing.py
Akshat412: Again, I'd rather add EPS warnings to the TM storage in a different PR.

flight/tasks/eps.py
```diff
@@ -64,16 +64,35 @@ class Task(TemplateTask):
...
+    }
+    log_counter = 0
 
     def __init__(self, id):
         super().__init__(id)
         self.name = "EPS"
 
-    def read_vc(self, sensor, voltage_idx, current_idx):
+    def read_vc(self, sensor):
         # log power monitor voltage and current
```
Akshat412, line 84: This comment is adding some confusion. I can see that read_vc and log_vc are basically just the old read_vc decoupled from logging this data, but it's not very obvious at a first glance.

flight/tasks/eps.py
```diff
@@ -87,6 +106,35 @@ def read_fuel_gauge(self):
         self.log_data[EPS_IDX.BATTERY_PACK_TTF] = int(fuel_gauge.read_ttf())
         self.log_data[EPS_IDX.BATTERY_PACK_TEMPERATURE] = int(fuel_gauge.read_temperature())
 
+    # Update MAV and set alert to indicate that resource is consuming too much power
+    # TODO: add torque coil alerts?
+    def set_power_alert(self, power, idx, threshold):
```
Akshat412, line 111: Would it make more sense to move this function into the EPS application and return the status (whether or not it is above the threshold) as a boolean / error code?

Akshat412 (changes requested):

flight/tasks/eps.py
achstar: Potentially - would also make unit testing easier

achstar:
hardware tested, verified that warnings do get raised

Akshat412 (approved):

perrintong325 (approved):

PR #219, opened 2025-03-17
Payload Control & Integration

### LINKED ISSUE #107: Payload: Control Interface
Circuitpython interface to interface with the Payload (commands and responses)

### LINKED ISSUE #116: Payload: Jetson <-> Mainboard UART Communication

### LINKED ISSUE #201: Payload: TM frame definition
Telemetry frame definition for the Jetson payload (Jetson telemetry, status of experiment, status of images, etc.).

### LINKED ISSUE #220: Payload SIL integration
Explicit

### LINKED ISSUE #221: Payload Task Integration
Explicit

### LINKED ISSUE #272: CDH: fix DH crash bug durign clean-up procedures with non-persistent data process
Fix a DH bug causing a crash when checking for the circular buffer (in SD card clean-ups) with a non-persistent process (RAM only).

Ibrassow:
This PR contains the end-to-end architecture for payload control and interfacing. 
- Full interprocess communication support for complete SIL testing with specific Payload FSW hash
- Command definitions, responses, and fault handling 
- Telemetry processing 
- Communication protocol (encoding and decoding), including file transfer
- External interface for CDH
- Payload task 
- UART interface

## Checklist of functionalities (tested in emulation)
- Payload state logic 
- File transfer
- Telemetry
- Supported commands
- External interfaces 
- Image transfer between mainboard and Jetson
- UART interface

## Protocol explanation

Some write-ups I had for myself about the sequence of events that I'm copy-pasting (will clean up later)
1) MB requests a specific file (image, OD data, etc - we could have different command for them but it doesn’t matter now)
For now, will do most recent
- Payload ACK the command 
- Payload selects the appropriate file and put in the temp buffer folder (locking it for sending). Also, sets internal variable for file transfer (counter, processing, etc), all this done at the command level)
- If file cannot be found (image not available (= not captured) or OD doesn’t have any results), we return a corresponding error ACK instead (and will error out any further requests for packet)
2) MB requests next packet in the typical command frame: [cmd_id, pckt_nb]
- Note that the pckt_nb will be the same as the sequence number in the ensuing response 
- And no, no need to put this into a sequence number in MB outgoing packets. This is unnecessary overhead for no value at all. Every byte counts
3) Payload sends the requested packet from the file (we can request in any order, doesn’t matter for the payload)
- If MB has some issues with the packet, it can just re-ask for the packet. If all good, it write to the DH.
- If the Payload encounters some issues, it will send an error ACK
- Depending on the cases, MB can re-request or eventually cancel the whole process by doing nothing (no need for cancel message, Payload just copies the file in a temp folder and will overwrite it when a new request arises anyway)
4) At some point, MB will ask a pckt_nb > total_seq_count on the payload  (basically asking more than the size). Payload will then send a NO_MORE packet. This is the indication for the MB that it received everything it needed. If it needs (for some obscure reason) a particular packet, it can still request it with the pckt_nb until it requests another file, which will overwrite the considered file.

## HW Validation 

- Validated all implemented commands (including file transfer) in emulation with Payload FSW (IPC)
- UART validation: in progress, currently held-up for new board (need power control)


Akshat412:
Linting failed smh smh

Ibrassow:
@Akshat412 @kmkar31 @michelleaheo Let's get some reviews rolling already
@Akshat412 Linting is failing because of some disabled tasks that I'll re-enable only at the end for now.

flight/apps/payload/controller.py
```diff
@@ -0,0 +1,520 @@
...
+
+    @classmethod
+    def add_request(cls, request: ExternalRequest) -> bool:
+        if not isinstance(request, int) or request < ExternalRequest.NO_ACTION or request >= ExternalRequest.INVALID:
+            # Invalid request
+            logger.error(f"Invalid request: {request}")
+            # TODO: add if a request is already being processed
+            return False
+        cls.current_request = request
+        cls.timestamp_request = TPM.monotonic()
```
s026134, line 165: Can you add a logger.info for successful add_request and also log any useful information like type of request and maybe the timestamp?

flight/apps/payload/controller.py
```diff
@@ -0,0 +1,520 @@
...
+
+    # OD variables
+    od_status: ODStatus = None
+
+    # (re-)boot variables
+    must_re_attempt_boot = False
+    attempting_reboot = False
+
+    # Power control
+    # en_pin = None
```
s026134, line 109: Why is this commented out? Should add a TODO here if this is something you'll be using in the future

flight/apps/payload/controller.py
Ibrassow: Removed it as this caused big log spams for no benefit.

flight/apps/payload/controller.py
Ibrassow: Power pin for the new carrier board, not there yet

flight/apps/payload/uart_comms.py
```diff
@@ -0,0 +1,50 @@
...
+            cls._uart = None
+            cls._connected = False
+
+    @classmethod
+    def disconnect(cls):
+        cls._connected = False
+
+    @classmethod
+    def send(cls, pckt):
+        cls.uart.write(pckt)
```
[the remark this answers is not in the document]
Ibrassow: not bad

emulator/cubesat.py
```diff
@@ -284,9 +284,30 @@ def WATCHDOG_AVAILABLE(self) -> bool:
         """
         return self.key_in_device_list("WATCHDOG") and self.__device_list["WATCHDOG"].device is not None
 
+    @property
+    def PAYLOADUART(self):
+        """PAYLOAD_EN: Returns the payload enable object
```
achstar, line 289: outdated comment? Might benefit from being a bit more specific

flight/apps/adcs/ad.py
```diff
@@ -74,7 +74,7 @@ class AttitudeDetermination:
     """ SENSOR READ FUNCTIONS """
 
     # ------------------------------------------------------------------------------------------------------------------------------------
-    def read_sun_position(self) -> tuple[int, np.ndarray, np.ndarray]:
+    def read_sun_position(self):  # -> tuple[int, np.ndarray, np.ndarray]:
```
achstar, line 77: What's the reasoning behind commenting these out? Did something break?

emulator/cubesat.py
Ibrassow: Good catch, it's copy paste from previous board

flight/apps/adcs/ad.py
```diff
@@ -84,7 +84,7 @@ def read_sun_position(self) -> tuple[int, np.ndarray, np.ndarray]:
 
         return status, sun_pos_body, np.array(light_sensor_lux_readings) / PhysicalConst.LIGHT_SENSOR_LOG_FACTOR
```
Ibrassow, line 85: note to myself: make this a constant mult

flight/apps/adcs/ad.py
Ibrassow: Not supported on my python version and got added by accident to the commits. Will uncomment once ready to merge

achstar: Oop okok good to know

flight/apps/payload/controller.py
Akshat412: Make an issue for adding control for this into FSW once the new en_pin is available (pin initialization should also be in the HAL, en_pin should just get the SATELLITE.JETSON_PIN object)

flight/apps/payload/controller.py
```diff
@@ -0,0 +1,520 @@
...
+            logger.info("Payload UART communication interface injected.")
+
+        elif SATELLITE.BUILD == "SIL":
+            from apps.payload.ipc_comms import PayloadIPC
+
+            cls.communication_interface = PayloadIPC
+            cls._interface_injected = True
+            logger.info("Payload IPC communication interface injected.")
+
+        assert cls.communication_interface is not None, "Communication interface not injected. Cannot initialize controller."
```
Akshat412, line 134: Note that assertions are not supported in CPy (I still have no idea why lol). This error message will never be seen.

flight/apps/payload/controller.py
```diff
@@ -0,0 +1,520 @@
...
+
+    @classmethod
+    def run_control_logic(cls):
+        # Move this potentially at the task level
+        cls._now = TPM.monotonic()
+
+        # Check for requests
+        cls.handle_external_requests()
+
+        if cls.state == PayloadState.OFF:
```
Akshat412, line 250: I know you don't have a crazy amount of free time but I'd like someone to try and make a state diagram thingy for this, this feels extremely important and an SM will be much easier to analyze.

Akshat412 (changes requested):

flight/apps/payload/controller.py
```diff
@@ -0,0 +1,520 @@
...
+                    cls._now - cls.time_we_started_booting > cls.TIMEOUT_BOOT
+                ):  # we seemingly failed --> attempt again to turn on
+                    cls.turn_off_power()  # turn off the power line, just in case
+                    cls._switch_to_state(PayloadState.OFF)  # Switch to OFF state
+                    cls.last_error = ErrorCodes.TIMEOUT_BOOT  # Log error
+                    cls.time_we_started_booting = 0  # Reset the boot time
+                    # CDH / HAL notification
+                    cls.must_re_attempt_boot = True  # Set the flag to re-attempt booting
+
+        elif cls.state == PayloadState.READY:
```
Akshat412, line 292: When do we go from PayloadState.READY to PayloadState.SHUTTING_DOWN? Is turning off the payload only done using an ExternalRequest currently?

Akshat412 (changes requested):

flight/apps/payload/controller.py
Ibrassow: this will be asserted in the SIL at least, as it should

flight/apps/payload/controller.py
Ibrassow: Any global state != Experiment

flight/apps/payload/controller.py
Ibrassow: It's not needed now. 1. There is nothing complicated, and it's a completely linear "power state of the Jetson". 2. Nothing is frozen until I freeze it. :) External docs come afterwards. 3. The code and comments are the source of truth. 
image

This comment also applies to the rest of the docs (global SM, telemetry, comms, etc.). Databases are good for quick look-ups after maturation, but code changes faster than documentation.

flight/apps/payload/controller.py
```diff
@@ -0,0 +1,520 @@
...
+                # We have timed out while booting
+                # Log error and reset the state
+                logger.error("Timeout booting. Resetting state.")
+                cls._switch_to_state(PayloadState.POWERING_ON)
+
+        elif cls.state == PayloadState.POWERING_ON:
+            # Wait for the Payload to be ready
+            cls.turn_on_power()
+
+            if cls.time_we_started_booting == 0:
```
Akshat412, line 265: If 0 is being used as a special value here to indicate cls.time_we_started_booting is not set, define it as a macro.

flight/apps/payload/controller.py
```diff
@@ -0,0 +1,520 @@
...
+            if not cls._did_we_send_a_command():  # Add timeout for retry
+                # Check for telemetry
+                if cls._now - cls._prev_tm_time > cls.telemetry_period and not cls._not_waiting_tm_response:
+                    cls.request_telemetry()
+
+            if not cls._did_we_send_a_command():  # Add timeout for retry
+                if FileTransfer.in_progress:
+                    # Check if we need to request the next file packet
+                    if not cls.just_requested_file_packet:
+                        cls.request_next_file_packet()
```
Akshat412, line 306: Same if condition is there twice? Why?

Akshat412 (dismissed):

flight/apps/payload/controller.py
Ibrassow: On purpose. Need to re-check that we sent a command (can't have both at the same time - well, payload supports it but the mainboard constraints (because of cp) don't play nice with it)

flight/apps/payload/controller.py
Ibrassow: This is not a special value but actually 0 sec :) . `cls.time_we_started_booting` is a monotonic timer while we're booting. It's being reset to zero once the payload is on.

flight/apps/payload/controller.py
Akshat412: Ok, sounds good. Add a comment specifying this.

flight/apps/telemetry/packing.py
```diff
@@ -502,7 +495,22 @@ def pack_tm_storage(cls):
...
-        pass
+        if not cls._TM_AVAILABLE:
+            cls._TM_AVAILABLE = True
+
+        cls._FRAME = bytearray(_TM_PAYLOAD_SIZE + 4)  # pre-allocated buffer for packing
+        cls._FRAME[0] = const(0x04) & 0xFF  # message ID
+        cls._FRAME[1:3] = pack_unsigned_short_int([const(0x00)], 0)  # sequence count
+        cls._FRAME[3] = const(_TM_PAYLOAD_SIZE) & 0xFF  # packet length
+
+        if DH.data_process_exists("payload_tm"):
```
Akshat412, line 506: Is this being done in this PR or another one?

flight/apps/payload/controller.py
Akshat412: Ok but this is still using 0 to indicate the timer needs to be set again...

Ibrassow:
It is currently in testing. This integration branch will also implement a specific FSW hash for the Payload CI to protect against regressions.

prchall123 (changes requested):
The SIL results shouldn't be pushed to the git history. I don't think the reconstruct_tilepack.py should be in scripts. Having the .bin files in tests isn't ideal, but I can approve if there is no other way.

ChaseDunaway21:
I have the same comments as Pedro, though I am concerned about how old some of these changes are. One thing I saw that was strange was that the UART Baudrate was set to 460800; was this from trial and error with the common rates?

prchall123 (approved):

ChaseDunaway21 (approved):

ChaseDunaway21 (approved):
This looks good, though the 'PAYLOADUART' naming convention doesn't seem to be the same everywhere. For instance,

flight/hal/argus_v4.py:316:        self.__payload_uart = ArgusV4Interfaces.JETSON_UART

and in flight/hal/argus_v1 it is "PAYLOAD_UART"

This is really minor but just something I saw in case it comes up later.

PR #225, opened 2025-03-18
Time Processing Module

Akshat412:
The Time Processing Module (TPM) acts as a controller for RTC access, returning RTC time when requested and updating the RTC's time when requested.

If the RTC fails, the TPM performs best-effort timekeeping. For this, it uses a known time reference to calculate an offset between "real time" (whose accuracy depends on the accuracy of the time reference) and time.time() from CircuitPython's firmware (which is always initialized to Jan 1st 2020). 

In case of RTC failures, this offset is added to time.time() and returned. This offset is also calculated every time the spacecraft receives a new time reference. Thus, RTC failures only lead to a slightly worse time reference onboard.

This PR adds:
- TPM in the FSW core
- TPM time access to all apps, core components and tasks that earlier used time.time()
- Uplinked time references can update the RTC time
- Error handling (as above) if the RTC fails
- Reworked commanding task with support for TPM initialization (also added things into functions)
- Reworked GPS task for more debugging information

Updating RTC time using GPS time references will be handled in a later PR (this one already got pretty big), as this needs some testing with valid fixes (outside). I also need to check the accuracy of GPS time as it correlates to fix modes.


Akshat412:
Need to fix the TPM to work with the simulator (need a working RTC driver).

flight/core/time_processor.py
```diff
@@ -0,0 +1,103 @@
...
+7 - The TPM offset shall also be updated for state correction using
+GPS time or an uplinked time reference.
+"""
+
+import time
+
+from hal.configuration import SATELLITE
+
+
+class TimeProcessor:
```
Ibrassow, line 36: as we want ideally the whole SC (and other people) to use it, name it easily with something like "Time" so you do Time.time()

flight/hal/cubesat.py
```diff
@@ -64,7 +64,7 @@ def __init__(self):
 
         self.__imu_temp_flag = False
 
-        # Debugging
+        # TODO: Is this even needed?
         self._time_ref_boot = int(time.time())
```
Ibrassow, line 68: You want to know your uptime in the TM, you need to know when you booted for that. It's better to use monotonic here though since we'll be computing always a difference

flight/tasks/gps.py
```diff
@@ -1,14 +1,23 @@
...
 
+"""
+    Fix Modes in GPS NMEA Message:
+
+    0 - No fix: The GPS receiver has not obtained a valid fix or has lost the fix.
+    1 - GPS fix: A 2D fix is available (latitude and longitude, but altitude is not necessarily reliable).
+    2 - Differential GPS fix: A 2D fix is available, with differential corrections applied for improved accuracy.
+    3 - 3D fix: A 3D fix is available (latitude, longitude, and altitude are all valid and reliable).
+    4 - GNSS fix: A fix using signals from multiple GNSS systems (e.g., GPS, GLONASS, Galileo, BeiDou).
+    5 - Time fix: A fix based on time synchronization, typically used in high-precision or timing applications.
```
Ibrassow, line 18: nice

flight/tasks/command.py
```diff
@@ -33,33 +33,63 @@ class Task(TemplateTask):
...
+
+        # NOTE: TPM time reference initialization
+        # In case the RTC has died, TPM uses time reference for time keeping
+        # The time reference is used to get offset from time.time()
+
+        # If an old time reference is not available, we depend on state correction
+        # from GPS or uplinked commands, and until then the time will be egregiously wrong
+
+        # This will not work until OBDH initializes CDH data process, so try 4 times
+        if (self.set_boot_cnt) < 4 and (SATELLITE.RTC_AVAILABLE is False):
```
Ibrassow, line 66: I think we need to properly set the STARTUP timeout (using actually our current uptime/boottime) in terms of time rather than boot count. We plan on doing the deployment in STARTUP too..

flight/tasks/command.py
```diff
@@ -71,76 +101,91 @@ async def main_task(self):
...
+        # Periodically log to serial
         self.log_print_counter += 1
+
         if self.log_print_counter % self.frequency == 0:
             self.log_print_counter = 0
-            self.log_info(f"Time: {int(time.time())}")
-            self.log_info(f"Time since boot: {int(time.time()) - SATELLITE.BOOTTIME}")
+
+            if not SATELLITE.RTC_AVAILABLE:
+                self.log_warning("RTC FAILURE: Time reference is best-effort from TPM")
```
Ibrassow, line 186: You don't need to log that warning after the first one. We'll already have the hardware failure (although I assume this is for debugging right now. Log messages must have a good S/N ratio too to be useful and efficient

Ibrassow:
Good. To get back to my Time.time() point, I think it might be best to try to replicate as much as possible the interface of the original time module (even with obvious wrappers like monotononic).

For the SIL side, creating the RTC object is straightforward. I'd suggest syncing with @kmkar31 as this directly influence how we're going to do the faster-than-realtime SIL

flight/hal/cubesat.py
Akshat412: Yeah this is fair, I can use a monotonic clock here. My thought for putting the boot_time in the commanding task init is that gives us our uptime too, but the start point is the SW's start time, not the HW's start time. If we need the HW start time to be the beginning of our boot time I can def do that.

flight/core/time_processor.py
Akshat412: TIme.time() feels too close to time.time(). Can't we stick with TimeProcessor lol, it adds some determinism in which one you end up using (less bugs)?

flight/tasks/command.py
Akshat412: Oh this is just a counter for initializing the TPM, we still have a legit boot time, it's just SW boot time rather than HW boot time.

flight/core/time_processor.py
Akshat412: I can call this TPM if the size of the name is too long.

flight/tasks/command.py
```diff
@@ -33,33 +33,63 @@ class Task(TemplateTask):
...
+            else:
+                # If no RTC or old time reference available, TPM goes back to init for time.time()
+                self.log_warning("Cannot set time reference as CDH process does not exist")
+
+            self.set_boot_cnt += 1
+
+        else:
+            # Set time reference
+            if self.set_boot_flag is False:
+                self.boot_time = TimeProcessor.time()
```
Akshat412, line 88: @Ibrassow this is the new "boot time" that we consider, which is then used for time since boot, and can be used for all timing related things including deployment. It's just the boot time for the software rather than the hardware, so does not accommodate the 1-2s of HW boot time. I can bring back SATELLITE.BOOTTIME with the monotonic clock.

perrintong325:
pystack exhausted is crazy

flight/core/time_processor.py
Ibrassow: Being close to time.time is the point actually. "Timing" seems better

flight/hal/cubesat.py
Ibrassow: Hmm good point, I think we might need both in this case. It's fair to call time.monotonic (not your module) in HW initialization and just have your Timing module having a wrapper for the rest of the software

flight/core/time_processor.py
Akshat412: Changed all access to this to TPM (in other tasks). TimeProcessor is always imported as TPM in all FSW now. Is that good?

flight/hal/cubesat.py
Akshat412: Done. Also added support in TPM for time.monotonic(), so FSW needn't import time anywhere. Time since boot in CDH uses this as a reference again.

Akshat412:
Verified that nothing in FSW is using time.time() (these are all comments)
image

flight/core/time_processor.py
```diff
@@ -0,0 +1,107 @@
...
+    # Initialize to 0 (no offset)
+    time_offset = 0
+
+    """
+        Name: calc_time_offset
+        Description: Calculate offset (time_reference - time.time())
+    """
+
+    @classmethod
+    def calc_time_offset(self):
```
kmkar31, line 61: Use ```cls``` for class methods instead of ```self```

flight/core/time_processor.py
```diff
@@ -0,0 +1,107 @@
...
+    """
+
+    @classmethod
+    def set_time(self, unix_timestamp):
+        if SATELLITE.RTC_AVAILABLE:
+            # RTC exists, update RTC time
+            SATELLITE.RTC.set_datetime(time.localtime(unix_timestamp))
+
+            # Update TPM time reference and offset, in case RTC fails later
+            self.time_reference = unix_timestamp
```
kmkar31, line 76: nit: Probably don't have code duplication. Since setting time reference and calculating offset are done on both conditions

flight/tasks/gps.py
```diff
@@ -42,23 +51,22 @@ def __init__(self, id):
...
 
+                # Check if the module sent a valid nav data message
                 if SATELLITE.GPS.update():
-                    # Assuming we have a fix for now
+                    # Check if the fix is at least a 2D fix (fix_mode >= 1)
                     if SATELLITE.GPS.has_fix():
-                        # TODO GPS frame parsing - get ECEF in (cm) and ECEF velocity in cm/s
-                        # TODO : Change the time to get the GPS time rather than the system time here
-                        self.log_info("GPS module got a fix")
+                        # If a valid message and a fix, update GPS log info
```
kmkar31, line 67: Shouldn't there be logic here to update the RTC's time based on the GPS fix time?

flight/core/data_handler.py
```diff
@@ -309,7 +309,7 @@ def create_new_path(self) -> str:
             str: The new filename.
         """
         # Keeping the tag name in the filename for identification in debugging
-        return join_path(self.dir_path, self.tag_name) + "_" + str(int(time.time())) + ".bin"
+        return join_path(self.dir_path, self.tag_name) + "_" + str(int(TPM.time())) + ".bin"
```
kmkar31, line 312: Are we sure two files will not be created within a second leading to one overwriting the other?

flight/core/time_processor.py
```diff
@@ -0,0 +1,107 @@
...
+    """
+        Name: set_time
+        Description: Set RTC time
+    """
+
+    @classmethod
+    def set_time(self, unix_timestamp):
+        if SATELLITE.RTC_AVAILABLE:
+            # RTC exists, update RTC time
+            SATELLITE.RTC.set_datetime(time.localtime(unix_timestamp))
```
kmkar31, line 73: Maybe check if the reference is more than a second from the current timestamp and only do a write cycle iff there is a large time difference (Otherwise, maintain a software offset?)

kmkar31 (changes requested):

flight/core/time_processor.py
Akshat412: Good catch

flight/tasks/gps.py
Akshat412: GPS time references are a little funny. I recently learnt that depending on the quality of your fix, the time can just straight up be wrong. I need to experiment outside with GPS data to see what level of fix will do this successfully. I'd rather handle GPS time reference updates in a different PR.

flight/core/data_handler.py
Akshat412: No, but we'd need a more accurate clock (like something from the time module) to distinguish beyond a second level for this stuff.

Akshat412: @Ibrassow any thoughts here? How likely are file creations within a second?

flight/apps/adcs/ad.py
```diff
@@ -92,7 +91,7 @@ def read_gyro(self) -> tuple[int, int, np.ndarray]:
 
         if SATELLITE.IMU_AVAILABLE:
             gyro = np.array(SATELLITE.IMU.gyro())
-            query_time = int(time.time())
+            query_time = int(TPM.time())
```
Ibrassow, line 94: can you just straight up return time in int directly from TPM.time()? so we can remove all those conversions

flight/core/data_handler.py
Ibrassow: Files get opened/created and written to only when you first log as you can see in resolve_current_file(), which never happens in STARTUP.

Also, the DH will reopen the latest file to append to it first (unless the corresponding flag is false). Only when the data process didn't exist or the current file was full will it generate a new path.

flight/core/logging.py
```diff
@@ -174,7 +175,7 @@ def format(self, record: LogRecord) -> str:
             "args": record.args,
         }
         if "{asctime}" in self.fmt or "%(asctime)s" in self.fmt:
-            now = time.localtime()
+            now = time.localtime(TPM.time())
```
Ibrassow, line 178: make a TPM.localtime() instead

flight/core/time_processor.py
```diff
@@ -0,0 +1,110 @@
...
+
+class TimeProcessor:
+    """
+    time_reference: Last known TPM time, either from
+    the last RTC reading or from the commanding task
+    (on boot) if the RTC has failed.
+    """
+
+    # Initialize to UTC timestamp for Jan 1st 2020 (time.time() init val)
+    time_reference = time.time()
```
Ibrassow, line 44: Are you sure this always gets to Jan 1st 2020 ?

flight/core/time_processor.py
```diff
@@ -0,0 +1,110 @@
...
+
+            # Return RTC time
+            return rtc_time
+        else:
+            # RTC has failed, return time.time() + TPM offset
+            return time.time() + cls.time_offset
+
+    @classmethod
+    def monotonic(cls):
+        return time.monotonic()
```
Ibrassow, line 110: return int directly (if not already the case) so we don't need to do it in the codebase

flight/tasks/command.py
```diff
@@ -33,33 +33,56 @@ class Task(TemplateTask):
...
+                    self.tpm_init_tries_cnt = 4
+                else:
+                    # If no RTC or old time reference available, TPM goes back to init for time.time()
+                    self.log_warning("Cannot set time reference as CDH process has no latest data")
+            else:
+                # If no RTC or old time reference available, TPM goes back to init for time.time()
+                self.log_warning("Cannot set time reference as CDH process does not exist")
 
-            # TODO: Burn wires
+            self.tpm_init_tries_cnt += 1
```
Ibrassow, line 80: use a time-based approach with monotonic (basically a timeout) rather than trying X amount of time, so it's independent of the task frequency

flight/tasks/command.py
```diff
@@ -71,76 +94,88 @@ async def main_task(self):
...
+            if queue_error_code == QUEUE_STATUS.OK and cmd_args != processor.CommandProcessingStatus.ARGUMENT_UNPACKING_FAILED:
+                self.log_info(f"Processing command: {cmd_id} with args: {cmd_args}")
+                status, response_args = processor.process_command(cmd_id, *cmd_args)
+                processor.handle_command_execution_status(status, response_args)
+
+                # Log the command execution history
+                self.log_commands[0] = int(TPM.time())
+                self.log_commands[1] = cmd_id
+                self.log_commands[2] = status
+                DH.log_data("cmd_logs", self.log_commands)
```
Ibrassow, line 144: Retest the whole CDH logic with GS since you modified the whole task

flight/tasks/command.py
```diff
@@ -71,76 +94,88 @@ async def main_task(self):
...
+            # Execute state machine
+            self.state_machine_execution()
 
-                    # Log the command execution history
-                    self.log_commands[0] = int(time.time())
-                    self.log_commands[1] = cmd_id
-                    self.log_commands[2] = status
-                    DH.log_data("cmd_logs", self.log_commands)
+            # Run command processor
+            self.command_processor_execution()
```
Ibrassow, line 156: I think commands should come before the state machine execution.

flight/tasks/gps.py
Ibrassow: Note, if this was not obvious: the CDH should handle that (and check the fix quality) and use your existing unix_timestamp interface in the task. Also, we would want to update the time on a schedule (daily?) based on the RTC drift. Too frequent updates might lead to weird things I feel...

Ibrassow (changes requested):

flight/tasks/gps.py
Akshat412: Yeah that makes sense. You only get about a million write cycles on the RTC, so more than weird things, at a certain point you will just never be able to update the time again.

flight/core/data_handler.py
Akshat412: So, in practice, we do not need sub-second precision for filepaths?

flight/tasks/command.py
Akshat412: Did test this extensively, including during the ridge test, and this is also tested with all GS commands. Everything still works about the same it's just encapsulated into functions.

flight/core/time_processor.py
Akshat412: I think it's controlled in firmware so it can be configured, but at least for the v2 mainboards it's always that date. Regardless, this may change with the v3 mainboards or something, I should remove that comment.

flight/tasks/command.py
Akshat412: Done

Akshat412:
TODO: SIL for the RTC is still missing.

flight/core/time_processor.py
Ibrassow: Wouldn't remove the comment but rather explain. I'd prefer if we explicitly set the constant in firmware and call that specific global constant instead

flight/core/data_handler.py
Ibrassow: no

flight/core/logging.py
```diff
@@ -174,7 +175,7 @@ def format(self, record: LogRecord) -> str:
             "args": record.args,
         }
         if "{asctime}" in self.fmt or "%(asctime)s" in self.fmt:
-            now = time.localtime()
+            now = TPM.localtime(TPM.time())
```
Ibrassow, line 178: This should be only TPM.localtime() - do the call internally. This is calling twice the TPM for nothing

flight/core/time_processor.py
```diff
@@ -0,0 +1,114 @@
...
+Time Processing Module (TPM)
+############################
+
+This module controls acts as a software interface to the RTC and
+an alternate source for time-of-day in case of RTC failures.
+
+Requirements:
+1 - The TPM shall always and only use UTC time.
+
+2 - The TPM shall be the sole point of access to the RTC.
```
Ibrassow, line 11: Would add "single point of reference for time," which sits a level above the RTC (which merely stores time). This distinction is important when considering the entire spacecraft, as the payload derives its time reference from the mainboard, and thus the TPM.

flight/core/time_processor.py
```diff
@@ -0,0 +1,114 @@
...
+            # RTC has failed, return time.time() + TPM offset
+            return int(time.time() + cls.time_offset)
+
+    @classmethod
+    def monotonic(cls):
+        return time.monotonic()
+
+    @classmethod
+    def localtime(cls, unix_timestamp):
+        return time.localtime(unix_timestamp)
```
Ibrassow, line 114: do not take any argument, just call cls.time() internally. Nothing else will give you the timestamp

flight/hal/cubesat.py
```diff
@@ -64,8 +64,8 @@ def __init__(self):
 
         self.__imu_temp_flag = False
 
-        # Debugging
-        self._time_ref_boot = int(time.time())
+        # Get boot time from time.monotonic()
+        self._time_ref_boot = int(time.monotonic())
```
Ibrassow, line 68: same thing here, remove the int and make that conversion internal to monotonic()

flight/tasks/command.py
```diff
@@ -33,33 +33,61 @@ class Task(TemplateTask):
...
+                    self.tpm_init_tries_cnt = 4
+                else:
+                    # If no RTC or old time reference available, TPM goes back to init for time.time()
+                    self.log_warning("Cannot set time reference as CDH process has no latest data")
+            else:
+                # If no RTC or old time reference available, TPM goes back to init for time.time()
+                self.log_warning("Cannot set time reference as CDH process does not exist")
+
+            self.tpm_init_cnt += 1
+
```
Ibrassow, line 86: This still uses a counter, which should be removed. If your timeout is 5 seconds, the number of attempts doesn’t matter. Just use time checks—this is exactly what monotonic is for: tracking durations. On each task execution, check the elapsed time since the first attempt, and give up when the timeout expires. No need for time multiplexing or added complexity.

Example pseudo code

image

flight/tasks/command.py
```diff
@@ -33,33 +33,61 @@ class Task(TemplateTask):
...
+        else:
+            # Check time_since_boot
+            time_since_boot = int(TPM.monotonic()) - SATELLITE.BOOTTIME
```
Ibrassow, line 89: see comment about int()

Ibrassow (changes requested):

flight/core/time_processor.py
Akshat412: Yeah this is a good idea. I can look into the firmware for this.

flight/core/logging.py
Akshat412: Doing this internally is what caused the pystack exhausted error.

flight/core/time_processor.py
Akshat412: Again, in the logger's case, this is what caused pystack exhausted.

flight/hal/cubesat.py
Akshat412: This is in HAL. Not using TPM here.

kmkar31 (commented):
1. Remove redundant .uf2 file
2. Change time_reference in TPM to a hardcided date

kmkar31 (approved):

firmware/Mainboard v2/mpconfigboard.h
```diff
@@ -3,9 +3,7 @@
 #define MICROPY_HW_BOARD_NAME "Argus2"
 #define MICROPY_HW_MCU_NAME "rp2040"
 
-//#define MICROPY_HW_LED_STATUS (&pin_GPIO25)
-
-//#define MICROPY_HW_NEOPIXEL (&pin_GPIO17)
+#define CIRCUITPY_PYSTACK_SIZE 2048
```
Ibrassow, line 6: so this is the line haha

flight/core/logging.py
```diff
@@ -174,7 +175,7 @@ def format(self, record: LogRecord) -> str:
             "args": record.args,
         }
         if "{asctime}" in self.fmt or "%(asctime)s" in self.fmt:
-            now = time.localtime()
+            now = TPM.localtime()
```
Ibrassow, line 178: now this is clean :)

Ibrassow (approved):
Looks good now, good job

PR #228, opened 2025-03-20
ADCS: Timeout set on MEKF Initialization

kmkar31:
- Sets a Timeout on MEKF Initialization
    - For MEKF init to work, we need valid GPS, sun sensor and magneto readings (which may not happen)
    - If any one sensor fails, MEKF init will always fail and the AD algorithm is useless
    - Now, there is a timeout and the MEKF is force initialized to a unit quaternion
    - We take a hit on the MEKF converging, but better than nothing

Tested on HW and works

### DIFF
--- a/flight/apps/adcs/ad.py
+++ b/flight/apps/adcs/ad.py
@@ -55,6 +55,8 @@ class AttitudeDetermination:
     last_position_update_time = 0
     last_gyro_update_time = 0
     last_gyro_cov_update_time = 0
+    mekf_init_start_time = None
+    mekf_timeout = 30  # seconds TODO: Decide a timeout and change
 
     # Sensor noise covariances (Decide if these numbers should be hardcoded here or placed in adcs/consts.py)
     gyro_white_noise_sigma = 0.01  # TODO : characetrize sensor and update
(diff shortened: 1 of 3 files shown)


Ibrassow:
Would MC this in sim and observe the convergence to see if that makes sense

Ibrassow:
Also, don't we rather want a condition on the orbit prop rather than gps?

kmkar31:
Will check sim, but apparently the payload guys are doing the same thing and their sim converges. Ofc, they are also sampling their IMU at a much faster rate

kmkar31:
> Also, don't we rather want a condition on the orbit prop rather than gps?

Wdym a condition on GPS?

Ibrassow:
> > Also, don't we rather want a condition on the orbit prop rather than gps?
> 
> Wdym a condition on GPS?

kmkar31:
Good point. I fixed this on the GPS<->ADCS interface PR since it makes more sense there. But yes, For Initialization to work, orbit prop should have a valid position and not necessarily GPS

kmkar31:
MEKF Results: 
With TRIAD initialization:
attitude_estimate_error

Force Init to a unit quaternion:
attitude_estimate_error

Higher inital error but converges to the same 3 sigma bound

Converges even if I set a unit quaternion to start with. MEKF updates at 1 Hz

Ibrassow (approved):

flight/apps/adcs/ad.py
```diff
@@ -162,6 +164,16 @@ def initialize_mekf(self) -> int:
         - This function is not directly written into init to allow multiple retires of initialization
         - Sets the initialized attribute of the class once done
         """
+        current_time = int(time.time())
```
perrintong325, line 167: make sure this is updated down the road with the TPM

perrintong325 (approved):

PR #235, opened 2025-03-22
Argus v3 HAL + Automatic configuration detection + Burn wire drivers

perrintong325:
**This PR is only intended for Argus 3 bring up, not all features on the new board are fully supported**
HAL for Argus v3.
  - Missing RW and burn wire driver driver.
  - Defaulted coils to be always on
  - TODO: interface for efuses, faults and alerts(HAL monitor PR), power reset/toggles(HAL monitor PR), coil enable interface
 
Auto Config Detection
 - Board version and CPY version are automatically detected, eliminating constants on board.
 - code.py automatically deleted on new boards

Burn Wire Driver
 - translated test script into driver

Data Handler
 - Updated to check if we can write to SD card instead of simply checking if path exists, needed for CPY9
 
Task Constants
 - Class length helper function

DRV8235
 - Fixed IV data access
 - Unified IV access interface (same as adm1176)

EPS
 - Accessing torque drivers correctly

tested on Argus 2(CPY8), Argus 3 (CPY9)


firmware/README.md
```diff
@@ -63,18 +64,20 @@ Navigate one level up (i.e. circuitpython/ports/[vendor]).
 Ensure that build-ArgusX folder does not exist in this directory, or else the compilation might fail.
 Run the following command to compile the target firmware
 ```
-make BOARD=ArgusX # ArgusX is the folder name in the boards folder
+# ArgusX is the folder name in the boards folder
+# -jY is an optional argument, where Y is the number of threads available on your computer
+make BOARD=ArgusX -jY
```
Ibrassow, line 69: you can do -j$(nproc) so it uses all available cores on your system to compile

flight/hal/configuration.py
```diff
@@ -1,42 +1,42 @@
...
     SATELLITE = ArgusV2(debug=DEBUG_MODE)
 
+elif board_id == "Argus3":
+    from hal.argus_v3 import ArgusV3
+
+    SATELLITE = ArgusV3(debug=DEBUG_MODE)
+
 else:
-    raise ValueError(f"Invalid hardware version {HARDWARE_VERSION}")
+    raise ValueError(f"Invalid hardware version {board_id}")
```
Ibrassow, line 42: good idea for the automatic detection

flight/hal/argus_v3.py
```diff
@@ -0,0 +1,611 @@
...
+
+        self.append_device("NEOPIXEL", self.__neopixel_boot)
+        self.append_device("REACTION_WHEEL", self.__reaction_wheel_boot)
+
+        self.__payload_uart = ArgusV3Interfaces.JETSON_UART
+
+        # TODO: maybe make this nicer or something
+        ArgusV3Components.PERIPH_PWR_EN.value = True
+        ArgusV3Components.COIL_EN.value = True
+
```
Ibrassow, line 298: Shouldn't it be Device(boot_fct) ?

Also, any reason why the rest of the light sensors, coils, etc are not added to the list 
(unless I missed them) ? Cubesat class doesn't add them 
image

flight/hal/argus_v3.py
```diff
@@ -0,0 +1,611 @@
...
+                ArgusV3Components.SD_BAUD,
+            )
+
+            vfs = VfsFat(sd_card)
+            mount(vfs, ArgusV3Components.VFS_MOUNT_POINT)
+            path.append(ArgusV3Components.VFS_MOUNT_POINT)
+            return [vfs, Errors.NOERROR]
+        except Exception as e:
+            if self.__debug:
+                raise e
```
Ibrassow, line 544: can you add back the comments (either here or in cubesat) that the SD card must be initialized first?

flight/hal/argus_v3.py
```diff
@@ -0,0 +1,611 @@
...
+    def __reaction_wheel_boot(self, _) -> list[object, int]:
+        try:
+            # TODO: reaction wheel driver + boot sequence
+            rw = None
+            return [rw, Errors.NOERROR]
+        except Exception as e:
+            if self.__debug:
+                raise e
+            # TODO: reaction wheel errors
+            return [None, Errors.NOERROR]
```
Ibrassow, line 604: irrespective of the fact that it shouldn't exist, if there is no driver, remove it for this PR. When the driver is there, add it at that point

firmware/Mainboard v3/mpconfigboard.h
```diff
@@ -4,12 +4,19 @@
...
-#define MICROPY_HW_BOARD_NAME "Pimoroni Pico Plus 2"
+#define MICROPY_HW_BOARD_NAME "Argus3"
 #define MICROPY_HW_MCU_NAME "rp2350b"
 
-#define MICROPY_HW_LED_STATUS (&pin_GPIO25)
+#define MICROPY_HW_NEOPIXEL (&pin_GPIO0)
 
-#define CIRCUITPY_BOARD_I2C         (1)
-#define CIRCUITPY_BOARD_I2C_PIN     {{.scl = &pin_GPIO1, .sda = &pin_GPIO0}}
+#define DEFAULT_I2C_BUS_SDA (&pin_GPIO24)
```
Akshat412, line 12: Before I forget, can you increase the pystack for the v3 firmware too?

Akshat412 (changes requested):

flight/core/data_handler.py
```diff
@@ -737,7 +756,7 @@ def scan_SD_card(cls) -> None:
         Example:
             DataHandler.scan_SD_card()
         """
-        if not path_exist(_HOME_PATH):
+        if not cls.__can_write_to_path(_HOME_PATH):
```
Akshat412, line 759: This was a really good catch, I was concerned about the v3 mainboards' response to a missing SD card lol

flight/core/logging.py
```diff
@@ -175,7 +175,7 @@ def format(self, record: LogRecord) -> str:
             "args": record.args,
         }
         if "{asctime}" in self.fmt or "%(asctime)s" in self.fmt:
-            now = TPM.localtime()
+            now = time.localtime(TPM.time())
```
Akshat412, line 178: Revert this change to match what main has. This was a super temporary fix I added one day while working on the v3 HAL , MB

flight/hal/argus_v3.py
```diff
@@ -0,0 +1,615 @@
...
+
+    # PERIPHERALS 3.3V
+    PERIPH_PWR_EN = digitalio.DigitalInOut(board.PERIPH_PWR_EN)
+    PERIPH_PWR_EN.direction = digitalio.Direction.OUTPUT
+    PERIPH_PWR_EN.value = True
+
+    PERIPH_PWR_OVC = digitalio.DigitalInOut(board.PERIPH_PWR_FLT)
+    PERIPH_PWR_OVC.direction = digitalio.Direction.INPUT
+
+    # DO NOT MOVE
```
Akshat412, line 32: Add a more in-depth comment here about the error you saw with some I2C devices having floating addresses, it's an important bug to log.

flight/hal/argus_v3.py
```diff
@@ -0,0 +1,615 @@
...
+    #########
+    # eFUSE #
+    #########
+
+    # PERIPHERALS 3.3V
+    PERIPH_PWR_EN = digitalio.DigitalInOut(board.PERIPH_PWR_EN)
+    PERIPH_PWR_EN.direction = digitalio.Direction.OUTPUT
+    PERIPH_PWR_EN.value = True
+
+    PERIPH_PWR_OVC = digitalio.DigitalInOut(board.PERIPH_PWR_FLT)
```
Akshat412, line 29: Why is PERIPH_PWR_OVC defined here while the other OVC GPIOs are defined in ArgusV3Components?

flight/hal/argus_v3.py
```diff
@@ -0,0 +1,615 @@
...
+class ArgusV3(CubeSat):
+    """ArgusV3: Represents the Argus V3 CubeSat."""
+
+    def __init__(self, debug: bool = False):
+        """__init__: Initializes the Argus V3 CubeSat."""
+        self.__debug = debug
+
+        super().__init__()
+
+        self.__payload_uart = ArgusV3Interfaces.JETSON_UART
```
Akshat412, line 310: Could use a brief explanation of why this field is set here for the v3 HAL only, and not the v2 HAL.

flight/hal/argus_v3.py
```diff
@@ -0,0 +1,615 @@
...
+        """
+        directions = {
+            "TORQUE_XP": [ArgusV3Components.TORQUE_XP_I2C_ADDRESS, ArgusV3Components.TORQUE_COILS_XP_I2C],
+            "TORQUE_XM": [ArgusV3Components.TORQUE_XM_I2C_ADDRESS, ArgusV3Components.TORQUE_COILS_XM_I2C],
+            "TORQUE_YP": [ArgusV3Components.TORQUE_YP_I2C_ADDRESS, ArgusV3Components.TORQUE_COILS_YP_I2C],
+            "TORQUE_YM": [ArgusV3Components.TORQUE_YM_I2C_ADDRESS, ArgusV3Components.TORQUE_COILS_YM_I2C],
+            "TORQUE_ZP": [ArgusV3Components.TORQUE_ZP_I2C_ADDRESS, ArgusV3Components.TORQUE_COILS_ZP_I2C],
+            "TORQUE_ZM": [ArgusV3Components.TORQUE_ZM_I2C_ADDRESS, ArgusV3Components.TORQUE_COILS_ZM_I2C],
+        }
+        # TODO: verify this driver actually works
```
Akshat412, line 422: Is this TODO still unverified?

Akshat412 (changes requested):

flight/hal/argus_v3.py
Ibrassow: @perrintong325 has https://github.com/cmu-argus-2/FSW-mainboard/pull/238 been fully tested?

flight/hal/cubesat.py
```diff
@@ -28,33 +28,34 @@ def __new__(cls, *args, **kwargs):
     def __init__(self):
         self.__device_list = OrderedDict(
             [
-                ("SDCARD", Device(self.__sd_card_boot)),
-                ("IMU", Device(self.__imu_boot)),
+                ("NEOPIXEL", Device(self.__neopixel_boot)),
+                ("SDCARD", Device(self.__sd_card_boot)),  # SD Card must enabled before other devices
```
Ibrassow, line 32: Include the full explanation (refer to the original pycubed class from Max for an example)

flight/hal/argus_v3.py
```diff
@@ -0,0 +1,615 @@
...
+                raise e
+
+            return [None, Errors.NEOPIXEL_NOT_INITIALIZED]
+
+    def reboot_device(self, device_name: str):
+        if device_name not in self.__device_list:
+            return Errors.INVALID_DEVICE_NAME
+        device = self.__device_list[device_name]
+        device.deinit()
+        # TODO: Implement reboot logic
```
Ibrassow, line 615: Can an issue be raised?

flight/hal/cubesat.py
```diff
@@ -28,33 +28,34 @@ def __new__(cls, *args, **kwargs):
...
+                ("TORQUE_ZP", Device(self.__torque_driver_boot)),
+                ("TORQUE_ZM", Device(self.__torque_driver_boot)),
+                ("LIGHT_XP", Device(self.__light_sensor_boot)),
+                ("LIGHT_XM", Device(self.__light_sensor_boot)),
+                ("LIGHT_YP", Device(self.__light_sensor_boot)),
+                ("LIGHT_YM", Device(self.__light_sensor_boot)),
+                ("LIGHT_ZM", Device(self.__light_sensor_boot)),
                 # ("LIGHT_ZP_1", Device(self.__light_sensor_boot)),
                 # ("LIGHT_ZP_2", Device(self.__light_sensor_boot)),
                 # ("LIGHT_ZP_3", Device(self.__light_sensor_boot)),
```
Ibrassow, line 61: There shouldn't be any comment but soft failures. If they don't exist in the children's classes, there should be void or None (or whatever mechanism you have)

flight/hal/drivers/pca9633.py
```diff
@@ -0,0 +1,60 @@
...
+        self._driver_enable = True
+
+    def disable_driver(self):
+        """Disable the PCA9633 driver."""
+        self._driver_enable = False
+        self._channel_enable = 0b00000000
+
+    def deinit(self):
+        """Deinitialize the PCA9633 driver."""
+        return
```
Ibrassow, line 60: Raise an issue for all the deinit

flight/main.py
```diff
@@ -24,9 +24,9 @@ def print_memory_stats(call_gc=True):
 
 print_memory_stats(call_gc=True)
 
-print("Booting ARGUS-1...")
+print("Booting ARGUS...")
```
Ibrassow, line 27: as it should have been a long time ago...

flight/tasks/adcs.py
```diff
@@ -264,10 +264,10 @@ def log(self):
...
         self.log_data[ADCS_IDX.LIGHT_SENSOR_YP] = int(self.AD.state[26]) & 0xFFFF
         self.log_data[ADCS_IDX.LIGHT_SENSOR_ZM] = int(self.AD.state[27]) & 0xFFFF
-        # self.log_data[ADCS_IDX.LIGHT_SENSOR_ZP1] = self.AD.state[28]
-        # self.log_data[ADCS_IDX.LIGHT_SENSOR_ZP2] = self.AD.state[29]
-        # self.log_data[ADCS_IDX.LIGHT_SENSOR_ZP3] = self.AD.state[30]
-        # self.log_data[ADCS_IDX.LIGHT_SENSOR_ZP4] = self.AD.state[31]
+        self.log_data[ADCS_IDX.LIGHT_SENSOR_ZP1] = int(self.AD.state[28]) & 0xFFFF
+        self.log_data[ADCS_IDX.LIGHT_SENSOR_ZP2] = int(self.AD.state[29]) & 0xFFFF
+        self.log_data[ADCS_IDX.LIGHT_SENSOR_ZP3] = int(self.AD.state[30]) & 0xFFFF
+        self.log_data[ADCS_IDX.LIGHT_SENSOR_ZP4] = int(self.AD.state[31]) & 0xFFFF
```
Ibrassow, line 270: @kmkar31 review those

flight/tasks/adcs.py
```diff
@@ -286,5 +286,9 @@ def log(self):
                 self.log_warning(msg)
             self.failure_messages = []
 
+            # Force read light sensors
+            self.AD.read_sun_position()
```
Ibrassow, line 290: @perrintong325  -  if this is done internally based on conditions, please remove from the PR as this might be unnecessary readings. Keep the PR focused on the HAL level changes.

Unless @kmkar31 agrees

flight/tasks/command.py
```diff
@@ -125,10 +133,19 @@ def state_machine_execution(self):
...
             pass
         elif SM.current_state == STATES.EXPERIMENT:
+            # Neopixel for EXPERIMENT (purple)
+            if SATELLITE.NEOPIXEL_AVAILABLE:
+                SATELLITE.NEOPIXEL.fill([255, 0, 255])
             pass
         elif SM.current_state == STATES.LOW_POWER:
+            # Neopixel for LOW_POWER (red)
+            if SATELLITE.NEOPIXEL_AVAILABLE:
+                SATELLITE.NEOPIXEL.fill([255, 0, 0])
```
Ibrassow, line 148: I understand why you might have done this but have you checked the computational delay incurred by those? There are going to be executing at each task run and by experience, neopixel doesn't play nice with time-sensitive computations

cc @Akshat412

Ibrassow (changes requested):
@kmkar31 @achstar Can you both review the whole PR as well? For such big changes, more eyes are preferrable

flight/hal/argus_v3.py
perrintong325: just forgot to remove ty

flight/tasks/adcs.py
perrintong325: he added them lol

flight/hal/argus_v3.py
```diff
@@ -0,0 +1,615 @@
...
+    # PERIPHERALS 3.3V
+    PERIPH_PWR_EN = digitalio.DigitalInOut(board.PERIPH_PWR_EN)
+    PERIPH_PWR_EN.direction = digitalio.Direction.OUTPUT
+    PERIPH_PWR_EN.value = True
+
+    PERIPH_PWR_OVC = digitalio.DigitalInOut(board.PERIPH_PWR_FLT)
+    PERIPH_PWR_OVC.direction = digitalio.Direction.INPUT
+
+    # DO NOT MOVE
+    time.sleep(2)  # Wait for peripherals to power up
```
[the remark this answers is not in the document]
perrintong325: mostly because of imu. ig we can maybe shorten the sleep time if we move the imu to the end of boot. Varun n I think 100ms should be more than enough for all the other peripherals to turn on, it's only there cos of the imu's clock stretch but yeah we can play with it

flight/hal/argus_v3.py
perrintong325: I mean it is, we just never merged a version of v2 Hal into main that has payload uart because... well technically v2 doesn't have payload uart it has a spi we just repurpose it. I don't think we need to explain that

flight/hal/drivers/pca9633.py
perrintong325: like an error? why would we do that? we are the ones calling deinit. deinit here is talking about deinitalising the pins

flight/tasks/command.py
Akshat412: This is one neopixel write every 0.5s I think we will be fine lol. I'll check task execution time for commanding with the logic analyzer again at some point today or tomorrow.

flight/core/logging.py
Akshat412: This isn't a merge issue. I hadn't updated the board I was using with the new firmware so the board had a smaller pystack. Temporary fix because this was during the vibe test assembly's bringup.

flight/hal/cubesat.py
Akshat412: This error is already handled. If a device doesn't exist it will be logged as an error in the boot sequence and the rest of FSW will boot normally.

The comments are primarily for the v2 mainboards for conserving RAM. Even empty CPy objects use too much RAM. For HAL v3 all devices can be enabled.

flight/apps/adcs/ad.py
```diff
@@ -81,6 +81,7 @@ def read_sun_position(self) -> tuple[int, np.ndarray, np.ndarray]:
         """
         light_sensor_lux_readings = read_light_sensors()
         status, sun_pos_body = compute_body_sun_vector_from_lux(light_sensor_lux_readings)
+        self.state[self.sun_lux_idx] = light_sensor_lux_readings
```
kmkar31, line 84: This was added just for vibe testing. Should be removed

flight/tasks/adcs.py
kmkar31: This stuff's fine. However, we need to scale light sensor readings down so that the max value <= 65535. This can be handled in #156  though

flight/tasks/adcs.py
kmkar31: I'd still prefer we added this locally and not push this to main

flight/tasks/adcs.py
```diff
@@ -286,5 +286,9 @@ def log(self):
...
+
             # Log Gyro Angular Velocities
             self.log_info(f"Gyro Ang Vel : {self.log_data[ADCS_IDX.GYRO_X:ADCS_IDX.GYRO_Z + 1]}")
+            self.log_info(f"Sun Sensors : {self.log_data[ADCS_IDX.LIGHT_SENSOR_XP:ADCS_IDX.LIGHT_SENSOR_ZM + 1]}")
```
kmkar31, line 294: This lux reading logs too should not go to main

simulation
kmkar31, line 1: Who reviewed the DRV check PR? It seems simulation was updated to an older commit. This should fix it though

flight/hal/drivers/pca9633.py
Akshat412: @Ibrassow Do you mean a GitHub issue lol

flight/tasks/eps.py
```diff
@@ -165,25 +163,45 @@ async def main_task(self):
...
+                    elif location == "YP":
+                        voltage, current = self.read_vc(sensor)
+                        self.log_vc("YP", EPS_IDX.YP_SOLAR_CHARGE_VOLTAGE, EPS_IDX.YP_SOLAR_CHARGE_CURRENT, voltage, current)
+                    elif location == "YM":
+                        voltage, current = self.read_vc(sensor)
+                        self.log_vc("YM", EPS_IDX.YM_SOLAR_CHARGE_VOLTAGE, EPS_IDX.YM_SOLAR_CHARGE_CURRENT, voltage, current)
+            for location, sensor in SATELLITE.TORQUE_DRIVERS:
+                if SATELLITE.TORQUE_DRIVERS_AVAILABLE(location):
+                    if location == "XP":
+                        voltage, current = self.read_vc(sensor)
```
achstar, line 182: The read_vc function assumes that a read_voltage_current() function exists in the driver for the corresponding sensor. I think the DRV8235 driver has separate read_voltage and read_current functions right now so this call will fail

achstar (dismissed):
Needs modifications to DRV8235 driver (just combine read_voltage and read_current fns into read_voltage_current)

Akshat412 (approved):

kmkar31 (approved):
Looks good. This PR has activates the light sensor logging but does not scale them to lie within 0->65535. This will throw errors in space but will not cause errors in the lab because our lux readings are never that high

achstar (approved):
eps changes look good

Ibrassow:
> Looks good. This PR has activates the light sensor logging but does not scale them to lie within 0->65535. This will throw errors in space but will not cause errors in the lab because our lux readings are never that high

Please, try to avoid next time, even if the intention is to correct it in a subsequent PR. This is bad practice to let known issues merged to main (even if temporary)

PR #238, opened 2025-03-27
Integrated DRV8235 driver

atletea:
Added functionality to read voltage/current output, fixed previous issues, tested and verified on X+-/Y+- boards.


flight/hal/drivers/drv8235.py
```diff
@@ -85,13 +85,14 @@ def __init__(self, i2c_bus, address):
...
         self._int_vref = True
+        # TODO: check inv_r_scale and inv_r values
         self._inv_r_scale = 0x3
         self._inv_r = 82
         self._en_out = True
         # Clear all fault status flags
         self.clear_faults()
 
-        super().__init__()
+        # super().__init__()
```
Akshat412, line 95: If the class does not inherent anything, just remove this. Adds confusion.

flight/hal/drivers/drv8235.py
```diff
@@ -85,13 +85,14 @@ def __init__(self, i2c_bus, address):
         self._reg_ctrl = 0x3  # Sets to voltage regulation
```
Akshat412, line 85: Nitpick, but shouldn't the __init__ method be after class variable definitions?

flight/hal/drivers/drv8235.py
```diff
@@ -207,6 +208,16 @@ def set_throttle_raw(self, new_throttle_raw):
...
         return
 
+    def read_voltage(self):
+        voltage_index = self._vmtr
+        voltage = VoltageAdapter.index_to_voltage(self, voltage_index)
+        return voltage
+
+    def read_current(self):
+        current_index = self._imtr
+        current = current_index * (3.7 / 255)  # max 3.7A range
```
Akshat412, line 218: Magic numbers, would like a little more explanation here.

flight/hal/drivers/drv8235.py
```diff
@@ -221,14 +232,16 @@ def fault(self):
         faults = []
         if self._fault:
             faults.append(Faults.DESCRIPTOR[0])
-            if self._ocp:
+            if self._stall:
                 faults.append(Faults.DESCRIPTOR[1])
```
Akshat412, line 236: Fault descriptor indices are technically magic numbers here, can they be defined as consts?

Akshat412 (changes requested):
Nothing majorly off, good. Make sure you merge from main ASAP.

flight/hal/drivers/drv8235.py
Ibrassow: agree, please @atletea modify the order to avoid confusion

flight/hal/drivers/drv8235.py
Ibrassow: also, use an already precomputed 3.7/255 instead of doing the division every time the sensor are called (+ we have a ton of them)

flight/hal/drivers/drv8235.py
```diff
@@ -65,7 +65,7 @@ class Faults:
            resumes with voltage restoration
     """
 
-    DESCRIPTOR = ["FAULT", "OCP", "OVP", "UVLO", "TSD"]
+    DESCRIPTOR = ["FAULT", "STALL", "OCP", "OVP", "TSD", "NPOR"]
```
Ibrassow, line 68: Use constant integers instead of strings

Ibrassow (changes requested):

flight/hal/drivers/drv8235.py
```diff
@@ -218,17 +217,33 @@ def fault(self):
...
-                faults.append(Faults.DESCRIPTOR[3])
-            if self._ilimit:
-                faults.append(Faults.DESCRIPTOR[4])
+                faults.append("OCP")
+            if self._ovp:
+                faults.append("OVP")
+            if self._tsd:
+                faults.append("TSD")
+            if self._npor:
+                faults.append("NPOR")
```
Ibrassow, line 246: define const integers and use those, no dynamically generated strings

flight/hal/drivers/drv8235.py
```diff
@@ -40,6 +40,14 @@ def voltage_to_index(self, volts):
...
 
 
+class CurrentAdapter:
+    """Output current calculator"""
+
+    def index_to_current(self, index):
+        """Convert an index value to current value based on max 3.7A current."""
+        return index * (3.7 / 255)
+
+
```
Ibrassow, line 50: why create a separate class to house a function? Also precompute the division

Ibrassow: make those operations in the respective functions, no need to increase the function calls for nothing in drivers. Simply document properly within the function where you do those oeprations

Akshat412 (changes requested):

Akshat412 (approved):

perrintong325 (approved):

PR #241, opened 2025-03-28
Faster than real-time SIL

kmkar31:
Addresses the following:
- Speeds up SIL by 50x (github CI instances can only achieve 20x though)
- Increases SIL duration to 10 minutes (equivalent to 2 orbits at 20x)


Akshat412 (approved):

Akshat412 (approved):
All the code looks fine to me, but merge from main.

flight/main.py
```diff
@@ -1,3 +1,4 @@
+# Import this first
```
Ibrassow, line 1: ?

Ibrassow (commented):
Make sure the sim values makes sense with the acceleration, can check this the gps/orbit output for instance. We don't want a fast but wrong SIL

PR #248, opened 2025-04-03
Time Reference Updates from GPS

Akshat412:
Based on the fix mode, the GPS can update the time reference for the RTC or only the TPM software module.
Also, added time reference for the comms task.


flight/tasks/gps.py
```diff
@@ -91,11 +101,18 @@ async def main_task(self):
 
                         DH.log_data("gps", self.log_data)
 
+                        # Internal reference update for TPM, regardless of fix quality
+                        TPM.set_time_reference(SATELLITE.GPS.unix_time)
```
kmkar31, line 105: I don't think you should update on a low fix mode at all. Moreover, you shouldn't let gps task decide when to update the rtc. Let the ```set_time_reference``` method decide on whether or not to write to the RTC depending on a calculated drift

kmkar31 (changes requested):

flight/tasks/gps.py
```diff
@@ -91,11 +101,19 @@ async def main_task(self):
 
                         DH.log_data("gps", self.log_data)
 
+                        # If RTC is inactive, set the TPM time reference regardless of fix quality
+                        if SATELLITE.RTC_AVAILABLE is False:
```
kmkar31, line 105: Yo, ``` if not SATELLITE.RTC_AVAILABLE```

kmkar31 (approved):

perrintong325 (approved):

sa2415 (approved):

s026134 (approved):

PR #251, opened 2025-04-07
CDH: Global State Machine for Argus

Akshat412:
Globalized state machine implementation for Argus' currently planned states:
- Adds all transitions in state machine (fig. below)
- FSW logs have label for all state transitions
- All state transitions based on modes from ADCS (angular velocity) and EPS (battery SoC)

Argus State Machine drawio

Additionally, for HAL v2, all peripherals not present on the MB are removed. This reduces RAM use for these boards by around 5%. Not everyone in the lab has v3 MBs yet so it's something we need to do (main now exceeds RAM threshold for v2 mainboards).

### DIFF
--- a/flight/tasks/command.py
+++ b/flight/tasks/command.py
@@ -168,7 +168,7 @@ class Task(TemplateTask):
             if self.EPS_MODE == EPS_POWER_FLAG.LOW_POWER:
                 # T1.2: Low SoC, transition to low power
                 self.log_info("T1.2: Transition from DETUMBLING to LOW POWER")
-                SM.switch_to(STATES.NOMINAL)
+                SM.switch_to(STATES.LOW_POWER)
 
         # ------------------------------------------------------------------------------------------------------------------------------------
         # NOMINAL
--- a/flight/tasks/command.py
+++ b/flight/tasks/command.py
@@ -68,15 +68,16 @@ class Task(TemplateTask):
         # Check time_since_boot
         time_since_boot = TPM.monotonic() - SATELLITE.BOOTTIME
 
-        # NOTE: TPM time reference initialization
-        # In case the RTC has died, TPM uses time reference for time keeping
-        # The time reference is used to get offset from time.time()
+        """
+        In case the RTC has died, the TPM uses a time reference for time keeping
+        The time reference is used to get offset from time.time() in CPy.
 
-        # If an old time reference is not available, we depend on state correction
-        # from GPS or uplinked commands, and until then the time will be egregiously wrong
+        If an old time reference is not available, we depend on state correction
+        from GPS or uplinked commands, and until then the time will be egregiously wrong.
 
-        # This will not work until OBDH initializes CDH data process, so try for 10 seconds
-        # since the SC has booted
+        This will not work until the OBDH initializes CDH data process, so we try this
+        for _TPM_INIT_TIMEOUT seconds since the SC has booted.
+        """
 
         if time_since_boot < _TPM_INIT_TIMEOUT and SATELLITE.RTC_AVAILABLE is False and self.time_ref_set is False:
             # Only worth it if the RTC is dead
@@ -112,7 +113,7 @@ class Task(TemplateTask):
 
                 # T0: Boot over and deployment complete
                 SM.switch_to(STATES.DETUMBLING)
-                self.log_info("Switching to DETUMBLING state.")
+                self.log_info("T0: Transition from STARTUP to DETUMBLING")
 
     def state_machine_execution(self):
         # ------------------------------------------------------------------------------------------------------------------------------------
--- a/flight/tasks/command.py
+++ b/flight/tasks/command.py
@@ -7,7 +7,8 @@ import gc
 import apps.command.processor as processor
 from apps.adcs.consts import Modes
 from apps.command import QUEUE_STATUS, CommandQueue
-from apps.telemetry.constants import ADCS_IDX, CDH_IDX
+from apps.eps.eps import EPS_POWER_FLAG
+from apps.telemetry.constants import ADCS_IDX, CDH_IDX, EPS_IDX
 from core import DataHandler as DH
 from core import TemplateTask
 from core import state_manager as SM
@@ -38,6 +39,10 @@ class Task(TemplateTask):
         super().__init__(id)
         self.name = "COMMAND"
         self.time_ref_set = False
+
+        # Transition status from ADCS and EPS
+        self.ADCS_MODE = Modes.STABLE
+        self.EPS_MODE = EPS_POWER_FLAG.NOMINAL
 
     def get_memory_usage(self):
         return int(gc.mem_alloc() / self.total_memory * 100)
@@ -105,6 +110,7 @@ class Task(TemplateTask):
                 if not DH.data_process_exists("cmd_logs"):
                     DH.register_data_process("cmd_logs", "LBB", True, data_limit=100000)
 
+                # T0: Boot over and deployment complete
                 SM.switch_to(STATES.DETUMBLING)
                 self.log_info("Switching to DETUMBLING state.")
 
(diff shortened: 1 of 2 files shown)


flight/tasks/command.py
```diff
@@ -113,40 +119,115 @@ def state_machine_execution(self):
...
+            eps_data = DH.get_latest_data("eps")
+
+            if eps_data:
+                self.EPS_MODE = eps_data[EPS_IDX.EPS_POWER_FLAG]
+            else:
+                self.log_warning("No latest EPS data available, assuming NOMINAL EPS mode")
+                self.EPS_MODE = EPS_POWER_FLAG.NOMINAL
+        else:
+            self.log_warning("EPS task not available, assuming NOMINAL EPS mode")
+            self.EPS_MODE = EPS_POWER_FLAG.NOMINAL
```
achstar, line 146: What is a scenario in which the EPS data process won't exist (if any)? Would it be better to assume a more conservative state here (i.e. low power)?

flight/tasks/command.py
```diff
@@ -113,40 +119,115 @@ def state_machine_execution(self):
...
+
         elif SM.current_state == STATES.LOW_POWER:
             # Neopixel for LOW_POWER (red)
             if SATELLITE.NEOPIXEL_AVAILABLE:
                 SATELLITE.NEOPIXEL.fill([255, 0, 0])
-            pass
+
+            if self.EPS_MODE != EPS_POWER_FLAG.LOW_POWER:
+                # T3.1: Nominal or high SoC, transition out of low power
+                self.log_info("T3.1: Transition from LOW POWER to NOMINAL")
```
achstar, line 208: Thinking about it more closely, some kind of hysteresis may be needed here. In a separate PR I'll add a moving average or something along those lines for the SOC and use that to update the eps power flag instead, but open to other thoughts too.

flight/tasks/command.py
Akshat412: So the reason I chose NOMINAL in this case instead of something more conservative here is for the possible scenarios where the EPS task is inactive: an extremely critical SW failure somewhere in the SC. At that point, choosing the NOMINAL power flag is essentially a Hail Mary to turn Argus into a beepsat for as long as we can.

But, this logic is not super thought out and can definitely have some issues. We can make this something more conservative like LOW_POWER.

flight/tasks/command.py
Akshat412: Given the entry and exit thresholds between NOMINAL and LOW_POWER, would this continue to be a problem? 

We would enter LOW_POWER at an SoC of 30% and exit at an SoC of 40%. So even if there is a condition in NOMINAL that is drastically draining our batteries, we would need to deplete 10% in total before going into LOW_POWER again.

image

achstar: Oh yeah I completely forgot about this. In that case yea this should be fine as-is

Akshat412: Good catch though! This would've generally been a problem, and I may need to have more descriptive comments about this.

flight/tasks/command.py
```diff
@@ -113,40 +119,115 @@ def state_machine_execution(self):
...
                 # Set the detumbling issue flag in the NVM
                 self.log_data[CDH_IDX.DETUMBLING_ERROR_FLAG] = 1
-                self.log_info("Switching to NOMINAL state after DETUMBLING timeout.")
+
+            if self.ADCS_MODE != Modes.TUMBLING or self.log_data[CDH_IDX.DETUMBLING_ERROR_FLAG] == 1:
+                # T1.1: Spin stabilized OR detumbling error flag is set (detumbling timeout)
+                self.log_info("T1.1: Transition from DETUMBLING to NOMINAL")
+                SM.switch_to(STATES.NOMINAL)
+
+            if self.EPS_MODE == EPS_POWER_FLAG.LOW_POWER:
```
perrintong325, line 168: technically this fall through is ok? I just don't know how I feel about it. because you are technically transitioning to nominal then transitioning again to low power, which a) doesn't match the state diagram and b) feels v redundant. maybe do if (T1.2) else if (T1.1)

flight/tasks/command.py
```diff
@@ -113,40 +119,115 @@ def state_machine_execution(self):
...
+                self.log_info("T1.2: Transition from DETUMBLING to LOW POWER")
                 SM.switch_to(STATES.NOMINAL)
 
+        # ------------------------------------------------------------------------------------------------------------------------------------
+        # NOMINAL
+        # ------------------------------------------------------------------------------------------------------------------------------------
+
         elif SM.current_state == STATES.NOMINAL:
-            # Neopixel for NOMINAL (orange)
+            # Neopixel for NOMINAL (green)
```
perrintong325, line 178: nice

flight/tasks/command.py
```diff
@@ -113,40 +119,115 @@ def state_machine_execution(self):
...
+
+            if self.ADCS_MODE != Modes.TUMBLING or self.log_data[CDH_IDX.DETUMBLING_ERROR_FLAG] == 1:
+                # T1.1: Spin stabilized OR detumbling error flag is set (detumbling timeout)
+                self.log_info("T1.1: Transition from DETUMBLING to NOMINAL")
+                SM.switch_to(STATES.NOMINAL)
+
+            if self.EPS_MODE == EPS_POWER_FLAG.LOW_POWER:
+                # T1.2: Low SoC, transition to low power
+                self.log_info("T1.2: Transition from DETUMBLING to LOW POWER")
                 SM.switch_to(STATES.NOMINAL)
```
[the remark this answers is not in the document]
Akshat412: Extremely good catch, thank you

flight/tasks/command.py
Akshat412: Is this referring to T1.2? I fixed it to resolve the previous comment, yeah technically the fall through would be ok but I'd want to stick to the state machine. Should be resolved now?

flight/tasks/command.py
```diff
@@ -113,40 +119,115 @@ def state_machine_execution(self):
...
+                # T2.1: High SoC, engage the payload
+                self.log_info("T2.1: Transition from NOMINAL to EXPERIMENT")
+                SM.switch_to(STATES.EXPERIMENT)
+
+            if self.EPS_MODE == EPS_POWER_FLAG.LOW_POWER:
+                # T2.2: Low SoC, transition to low power
+                self.log_info("T2.2: Transition from NOMINAL to LOW POWER")
+                SM.switch_to(STATES.LOW_POWER)
+
+            if self.ADCS_MODE == Modes.TUMBLING and self.log_data[CDH_IDX.DETUMBLING_ERROR_FLAG] != 1:
```
perrintong325, line 192: same issue will the fall throughs, they don't technically match the diagram. the two EPS transitions are understandable since they shouldn't both raise at the same time, but still a risk since we have to guarantee EPS to be always correct. Also once we transitioned to either of those two states you can still transition to detumble. same as above, maybe should prioritize detumbling

Akshat412: I see, should this be a bunch of if - elif statements to prevent this? I can def do that having a lot of independent if statements in here is not ideal.

flight/tasks/command.py
```diff
@@ -113,40 +119,115 @@ def state_machine_execution(self):
...
         if SM.current_state == STATES.DETUMBLING:
             # Neopixel for DETUMBLING (orange)
             if SATELLITE.NEOPIXEL_AVAILABLE:
                 SATELLITE.NEOPIXEL.fill([255, 165, 0])
 
-            # Check detumbling status from the ADCS
-            if DH.data_process_exists("adcs"):
-                if DH.get_latest_data("adcs")[ADCS_IDX.MODE] != Modes.TUMBLING:
-                    self.log_info("Detumbling complete - Switching to NOMINAL state.")
-                    SM.switch_to(STATES.NOMINAL)
```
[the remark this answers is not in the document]
Akshat412: #252

flight/tasks/command.py
Akshat412: Added a large header comment when these tasks are accessed, to address both assumptions this state machine makes.

Akshat412:
State machine is updated to match changes made to T2.X transitions (transitions out of NOMINAL).

kmkar31 (approved):
Will need payload flag to switch out of EXPERIMENT based on payload state

perrintong325 (dismissed):

achstar (approved):
Looks good to me

flight/tasks/command.py
```diff
@@ -105,48 +111,158 @@ def startup(self):
...
+        machine assumes a stable / nominal mode as a last resort "Hail Mary", to get
+        as much use out of the spacecraft as possible before it stops operation.
+        """
+
+        # Get ADCS mode (hysteresis management done in ADCS application)
+        if DH.data_process_exists("adcs"):
+            adcs_data = DH.get_latest_data("adcs")
+
+            if adcs_data:
+                self.ADCS_MODE = adcs_data[ADCS_IDX.MODE]
```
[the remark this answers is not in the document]
Akshat412: Good catch! Fixed this in the new commit

Akshat412:
Payload flag PR #253

achstar (approved):

perrintong325 (approved):

kmkar31 (approved):

PR #261, opened 2025-04-10
HAL controller

perrintong325:
- HAL
  - V3: included all reboot control and interface(power control).
  - All older versions: reboot using supervisor, no error handling.

- CubeSat class
  - new variables in device class to accommodate for reboot logic and mapped with associated power lines
  - "dead" variable to know if a device should be rebooted
  - ASIL(Argus Safety Integrity Level) to determine device's importance

- SD card
  - VFS wrapper to support deinit

- Firmware
    - Custom adafruit driver and firmware to support deinit

- Error List
  - New unified error list, driver level to be completed
 
tested on Argus3(all ASIL level triggered reboot, ensure "dead" device doesn't attempt to reboot, MCU hard reboot), Argus2(supervisor soft reboot)


build_tools/build-emulator.py
```diff
@@ -81,39 +80,6 @@ def create_build(source_folder, emulator_folder):
...
-def copy_folder(build_folder, destination_folder, show_identical_files=True):
-    for root, dirs, files in os.walk(build_folder):
-        for file in files:
-            source_path = os.path.join(root, file)
-            relative_path = os.path.relpath(source_path, build_folder)
-            destination_path = os.path.join(destination_folder, relative_path)
-
-            if not os.path.exists(os.path.dirname(destination_path)):
-                os.makedirs(os.path.dirname(destination_path))
-
```
Ibrassow, line 93: Can you keep changes focused on separate PRs? This is again the same thing about merging. Please avoid doing it in between branches, especially when unnecessary, as in there. It messes up the commit history, and reviews just become painful.

Ibrassow: You can just rebase in your next PR. Checkout a new branch from main and do your commits. Then whenever you want to sync, git rebase -i origin/main. If there is some conflicts, resolve them quick then git add . and git rebase --continue. Everything clean, your commits are placed on top of main

flight/hal/argus_v2.py
```diff
@@ -343,12 +342,12 @@ def __power_monitor_boot(self, location) -> list[object, int]:
...
             power_monitor = ADM1176(bus, address)
 
-            return [power_monitor, Errors.NOERROR]
+            return [power_monitor, Errors.NO_ERROR]
 
         except Exception as e:
             if self.__debug:
                 raise e
-            return [None, Errors.ADM1176_NOT_INITIALIZED]
+            return [None, Errors.DEVICE_NOT_INITIALISED]
```
Ibrassow, line 350: Why? I don't think this is the right move. Errors.DEVICE_NOT_INITIALISED tells you nothing to help debugging or for telemetry. I'd put back the device-specific initialization error

perrintong325: Because of the device class you don’t need it anymore. They are specific to a device, just simplified down to using the same constant. The new device class makes it even more specific than the old way of doing it because now we know the exact device that is throwing the error instead of just 1 error code that can come from any of the power monitors for example.

PR #262, opened 2025-04-10
Thermal: Battery Heaters

achstar:
Adding battery heater enable/disable logic to EPS.

Currently:

- Enable battery heaters if temperature drops below 0 degrees C and we are not in low power mode (?)
- Disable battery heaters if we enter low power mode or if temperature goes above 5 degrees C

My questions:
- Should we disable battery heaters if we enter low power? Also, is 5 degrees too small of a margin? In the datasheet, it says the charging operating temperature is 10-45 degrees C and discharging operating temperature is -20-60 degrees, so I wonder if I should raise the temperature threshold. 

Testing: unit tests + hardware testing in the lab by increasing temp thresholds and observing if heaters activated accordingly


flight/apps/eps/eps.py
```diff
@@ -9,13 +9,20 @@ class EPS_POWER_FLAG:
...
     LOW_POWER_ENTRY = const(30)
     LOW_POWER_EXIT = const(40)
     EXPERIMENT_ENTRY = const(80)
     EXPERIMENT_EXIT = const(60)
 
 
+# Temperature thresholds in cC (centi-Celsius)
+class EPS_TEMP_THRESHOLD:
+    BATTERY_HEAT_ENABLE = const(0)
+    BATTERY_HEAT_DISABLE = const(500)
```
perrintong325, line 23: where are these values coming from rn? is it datasheet or just placeholders

perrintong325:
tbh imo it's better safe than sorry especially if it's the battery. there's no point going into a low battery state that might end up damaging the battery

Ibrassow:
> Should we disable battery heaters if we enter low power? Also, is 5 degrees too small of a margin? In the datasheet, it says the charging operating temperature is 10-45 degrees C and discharging operating temperature is -20-60 degrees, so I wonder if I should raise the temperature threshold.

Hard to say without data. If only we had a proper thermal analysis... Your best bet is tuning with TVAC experiments. Be wary that our spacecraft can experience drastic temperature gradients, and it's not like we have an extensive set of thermal mitigation strategies (read: nothing)

> Should there be some entry in the EPS's DH buffer to indicate whether the battery heaters are enabled?

Yes

flight/apps/eps/eps.py
achstar: charging operating temperature for these batteries: 10-45 degrees C
discharging/storage operating temperature: -20-60 degrees C
https://www.dnkpower.com/wp-content/uploads/2022/08/SANYO-NCR18650GA-Datasheet.pdf
I chose something in the middle but in retrospect that doesn't make a whole lot of sense. I might just change the enable threshold to 10 C and the disable threshold to 15 C to ensure we're always within the charging operating temperature range (but then I'm concerned that we'll be enabling the heaters too often)

achstar:
hardware tested (but in the lab, so not really representative.) Confirmed that heaters do in fact turn on / off at given temp thresholds.

flight/apps/telemetry/constants.py
```diff
@@ -58,6 +58,7 @@ class EPS_IDX:
     ZP_SOLAR_CHARGE_CURRENT = const(40)
     ZM_SOLAR_CHARGE_VOLTAGE = const(41)
     ZM_SOLAR_CHARGE_CURRENT = const(42)
+    BATTERY_HEATERS_ENABLED = const(43)
```
achstar, line 62: Do you want me to add the associated packing.py changes in this PR @Akshat412 ?

flight/apps/eps/eps.py
Akshat412: We might want different entry and exit thresholds for both enable and disable to avoid hysteresis, the way we do for the EPS_POWER_FLAG.

flight/tasks/eps.py
```diff
@@ -251,6 +258,23 @@ async def main_task(self):
                     else:
                         self.log_error("EPS state invalid; SOC or power flag may be corrupted")
 
+                if SATELLITE.BATTERY_HEATERS_AVAILABLE:
```
Akshat412, line 261: The EPS main task could be a little more encapsulated IMO. Can you put this and some of the other operations in individual functions?

flight/apps/telemetry/constants.py
Akshat412: I'd say hold off, maybe just make an issue for it and assign me? I'd have to change the satellite and GS sides so it's easier done together. I can do it in this PR if that works.

Akshat412 (changes requested):

Akshat412:
> Should we disable battery heaters if we enter low power? Also, is 5 degrees too small of a margin? In the datasheet, it says the charging operating temperature is 10-45 degrees C and discharging operating temperature is -20-60 degrees, so I wonder if I should raise the temperature threshold.

The heaters consume 10W of power when turned on. However, they may not need to be on for super long. The most important context here is the fact that the batteries stop charging below 10*C, so staying cold in low power is essentially useless.

We should see how quickly internal temperature rises in the next TVAC and use that as an indicator for whether or not we enable them in low power or not.

flight/tasks/eps.py
achstar: Yep can do

achstar: Also in regards to the entry/exit threshold comment. Isn't that what I'm already doing? We enable heaters at 0 degrees C and disable at 5 degrees C (500 cC). Or do you think additional hysteresis is needed

flight/apps/eps/eps.py
achstar: Isn't that what I'm already doing? We enable heaters at 0 degrees C and disable at 5 degrees C (500 cC). Or do you think additional hysteresis is needed

flight/tasks/eps.py
Akshat412: This level of encapsulation is perfect, thanks!

flight/tasks/eps.py
```diff
@@ -136,16 +144,119 @@ def set_power_alert(self, voltage, current, idx, threshold):
...
+            self.log_data[EPS_IDX.BATTERY_HEATERS_ENABLED] = 1
+            self.log_info("Enabled battery heaters")
+        if SHOULD_DISABLE_HEATERS(enabled, temp, flag):
+            heaters.disable_heater0()
+            heaters.disable_heater1()
+            self.log_data[EPS_IDX.BATTERY_HEATERS_ENABLED] = 0
+            self.log_info("Disabled battery heaters")
+
+    def process_pm_readings(self, location, sensor):
+        if location == "BOARD":
```
Akshat412, line 163: This is missing the 3.3V peripheral line PM (mainboard v3s) / GPS power monitor (mainboard v2s). Since v3 is the main design now, I'd just call this the 3.3V peripheral power monitor. 

Just add a location for it in this function.

Akshat412 (changes requested):

flight/apps/eps/eps.py
```diff
@@ -9,16 +9,24 @@ class EPS_POWER_FLAG:
...
+
+
 # Power threshold in mW
 class EPS_POWER_THRESHOLD:
     MAINBOARD = const(1000)  # TODO: this threshold makes sense for v2 mainboards, but change to 400 for v3
+    PERIPH = const(1000)
```
Ibrassow, line 29: nit: `PERIPHERAL`

flight/tasks/eps.py
```diff
@@ -136,16 +145,107 @@ def set_power_alert(self, voltage, current, idx, threshold):
...
+
+    def process_fuel_gauge_readings(self):
+        self.log_info(f"Battery Pack Temperature: {self.log_data[EPS_IDX.BATTERY_PACK_TEMPERATURE]}°cC")
+        self.log_info(f"Battery Pack Reported SOC: {self.log_data[EPS_IDX.BATTERY_PACK_REPORTED_SOC]}% ")
+        self.log_info(f"Battery Pack Reported Capacity: {self.log_data[EPS_IDX.BATTERY_PACK_REPORTED_CAPACITY]} mAh ")
+        self.log_info(f"Battery Pack Current: {self.log_data[EPS_IDX.BATTERY_PACK_CURRENT]} mA ")
+        self.log_info(f"Battery Pack Voltage: {self.log_data[EPS_IDX.BATTERY_PACK_VOLTAGE]} mV ")
+        self.log_info(f"Battery Pack Midpoint Voltage: {self.log_data[EPS_IDX.BATTERY_PACK_MIDPOINT_VOLTAGE]} mV ")
+        self.log_info(f"Battery Pack Time-to-Empty: {self.log_data[EPS_IDX.BATTERY_PACK_TTE]} seconds ")
+        self.log_info(f"Battery Pack Time-to-Full: {self.log_data[EPS_IDX.BATTERY_PACK_TTF]} seconds ")
```
Ibrassow, line 228: To ease the logging load, you could maybe make a single string with all this and self.log_info at once. Each string is dynamically allocated

flight/apps/eps/eps.py
achstar: I'd like to keep this consistent with the HAL naming for the peripheral power (PERIPH_PWR_EN, etc). Tho open to change it, since the other PM names aren't totally consistent with the HAL names either

flight/tasks/eps.py
achstar: Looks like this now, is this alright? Or should I put some more effort into aligning it
Capture

achstar: We currently log battery pack readings every second. I'll go ahead and reduce it to every 5 seconds

Akshat412 (approved):

flight/tasks/eps.py
```diff
@@ -136,16 +146,109 @@ def set_power_alert(self, voltage, current, idx, threshold):
             elif idx == EPS_WARNING_IDX.ZM_COIL_POWER_ALERT:
                 self.log_warning(f"ZM Coil Avg Power Consumption Warning: {power_avg} with threshold {threshold} mW")
 
+    def set_battery_heaters(self, heaters):
+        enabled = heaters.heater0_enabled() or heaters.heater1_enabled()
+        temp = self.log_data[EPS_IDX.MAINBOARD_TEMPERATURE]
+        if SATELLITE.FUEL_GAUGE_AVAILABLE:
+            temp = self.log_data[EPS_IDX.BATTERY_PACK_TEMPERATURE]
```
Ibrassow, line 153: How much of a difference is there between the 2 temperature readings? How is the heater control strategy affected if we only have the MB temperature?

Ibrassow (commented):
Small renaming and a clarifying question, then it should be good to go (based on the answer :) )

flight/tasks/eps.py
Akshat412: MB temperature is typically higher by a few degree, we might want an offset to make the battery heater enabling a little more conservative

flight/apps/eps/eps.py
```diff
@@ -67,4 +71,14 @@ def GET_POWER_STATUS(buf, power, threshold, window):
...
+
+
+def SHOULD_ENABLE_HEATERS(enabled, temp, flag):
+    """returns whether battery heaters should be enabled in current conditions"""
+    return not enabled and flag != EPS_POWER_FLAG.NONE and (temp <= EPS_TEMP_THRESHOLD.BATTERY_HEAT_ENABLE)
+
+
+def SHOULD_DISABLE_HEATERS(enabled, temp, flag):
+    """returns whether battery heaters should be disabled in current conditions"""
+    return temp >= EPS_TEMP_THRESHOLD.BATTERY_HEAT_DISABLE
```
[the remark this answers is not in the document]
achstar: Hmm the only concern with that is, if these conditions are always true, we want the battery heaters to be on right? The thresholds are pretty conservative right now ; if the temperature goes below that, charging will be severely affected (we might not be able to charge as it will be well below the charging operating conditions of the batteries. Should confirm this in TVAC though.) So if we implement a timeout that expires and the batteries are still not at their desired temp, what can we do?

flight/tasks/eps.py
achstar: added 200 cC offset, not sure if that's too much but better to be conservative

Akshat412 (approved):

perrintong325 (approved):

PR #263, opened 2025-04-11
HAL controller

perrintong325:
- HAL
  - V3: included all reboot control and interface(power control).
  - All older versions: reboot using supervisor, no error handling.

- CubeSat class
  - new variables in device class to accommodate for reboot logic and mapped with associated power lines
  - "dead" variable to know if a device should be rebooted
  - ASIL(Argus Safety Integrity Level) to determine device's importance

- SD card
  - VFS wrapper to support deinit

- Firmware
  - Custom adafruit driver and firmware to support deinit

- Error List
  - New unified error list, driver level to be completed

tested on Argus3(all ASIL level triggered reboot, ensure "dead" device doesn't attempt to reboot, MCU hard reboot), Argus2(supervisor soft reboot)


flight/hal/drivers/adm1176.py
```diff
@@ -276,7 +276,7 @@ def __on_off_test(self) -> int:
             print("Error: Could not turn on device after turning off")
             return Errors.ADM1176_COULD_NOT_TURN_ON
```
kmkar31, line 277: ADM1176-specific errors don't seem to be defined yet

kmkar31 (commented):
What is actually monitoring for errors? I see the ```handle_error``` method to make reboot decisions but shouldn't there be a task to actually monitor errors and call this method?

flight/hal/drivers/adm1176.py
perrintong325: hal monitor will be in a diff pr because there's two different tasks(eps and Hal monitor) so need to merge this first

perrintong325: yeah that was in the old file forgot to move it to new one my b updated now.

kmkar31 (approved):
Is the task monitoring driver errors going to be in a new PR?

flight/tasks/obdh.py
```diff
@@ -25,19 +29,31 @@ async def main_task(self):
...
 
             if self.CLEANUP_COUNTER >= self.CLEANUP_COUNT_THRESHOLD:
                 DH.check_circular_buffers()
                 DH.clean_up()  # Clean up path that have been marked for deletion
                 self.CLEANUP_COUNTER = 0
 
+            if self.GC_COUNTER >= self.GC_COUNT_THRESHOLD:
+                gc.collect()
+                self.GC_COUNTER = 0
+
```
Ibrassow, line 56: GC is already called after each task execution cycle
image

perrintong325: @Akshat412 is a big fat liar

perrintong325:
> Is the task monitoring driver errors going to be in a new PR?

yes

kmkar31 (approved):

Ibrassow (approved):

Akshat412 (approved):

PR #268, opened 2025-04-16
SIL Improvements and Speedup

kmkar31:
Handles the following:
- Adds MAX17205 to the emulator
    - Can be used to toggle state transitions (by changing initial soc)
- Adds power monitors for solar panels and jetson
- Adds RTC


kmkar31:
Found some bug in GNC sim that causes RK4 to spit nan values after ~5 orbits

kmkar31:
1. Ground track on a 250x speedup 8 second real world runtime
ground_track

2. Ground track on a 125x speedup 14 second real world runtime
ground_track

The times are not exactly 2x because it takes about a second to copy all files to the build folder

Ibrassow:
Can you increase the SIL time to run 24h worth of orbits?

kmkar31:
> Can you increase the SIL time to run 24h worth of orbits?

Done

flight/core/states.py
```diff
@@ -53,7 +53,7 @@ class STATES:
         LOW_POWER: [NOMINAL, LOW_POWER],
     }
 
-    DETUMBLING_TIMEOUT_DURATION = 30  # seconds - TODO: Update with actual value
+    DETUMBLING_TIMEOUT_DURATION = 1.5 * 60 * 60  # seconds - TODO: Update with actual value
```
Ibrassow, line 56: where this value is coming from?

Ibrassow (dismissed):
Can you raise an issue for a config interface for the SIL on the MB side (without having to deal with the backend for that)?

perrintong325 (dismissed):

flight/apps/adcs/ad.py
```diff
@@ -467,7 +467,7 @@ def EKF_update(self, H: np.ndarray, innovation: np.ndarray, R_noise: np.ndarray)
         # S is scaled up by 1000 and then down to remove this error
         # If the singularity persists, we stop the covariance update
         try:
-            S_inv = 1000 * np.linalg.inv(1000 * S)
+            S_inv = np.linalg.inv(S)
```
Akshat412, line 470: Why is this multiplication changed (saying again cuz I think the last PR wasn't published correctly).

Akshat412 (changes requested):

Akshat412 (dismissed):

perrintong325 (dismissed):

flight/core/states.py
Ibrassow: > What's the reasoning behind this value?

Still haven't had an answer

Ibrassow:
> Can you raise an issue for a config interface for the SIL on the MB side (without having to deal with the backend for that)?

?

flight/core/states.py
kmkar31: Well, my analysis (from algorithm fixes on #276) says we detumble in like 40 min as long as we start under 20 deg/s. As a conservative value, I set it to 90 min since we never reattempt detumbling once the flag is set

kmkar31:
> Can you raise an issue for a config interface for the SIL on the MB side (without having to deal with the backend for that)?

Not sure I understand. What does a config for SIL mean?

Ibrassow (dismissed):

Ibrassow:
> > Can you raise an issue for a config interface for the SIL on the MB side (without having to deal with the backend for that)?
> 
> Not sure I understand. What does a config for SIL mean?

Configuration files that is read at the start to set all parameters of the SIL/simulation: run time, faults to inject, device availability, etc

PR #269, opened 2025-04-16
Comms: DOWNLINK_ALL CMD + File Time Fixes

Akshat412:
New state for the internal comms SM: DOWNLINK_ALL. If entered (a precondition for which is a valid filepath), comms will stay in this state and downlink all file packets for the file at the associated path.

Comms Statechart drawio

This is a worst-case scenario command, to be used only at the end of a ground pass or if we know of any critical errors and just want to get as much use out of the spacecraft as possible.

Additionally, this PR fixes some issues in finding files at a particular "file_time" as requested by the GS.

### DIFF
--- a/flight/apps/command/commands.py
+++ b/flight/apps/command/commands.py
@@ -137,7 +137,7 @@ def REQUEST_FILE_METADATA(file_id, file_time=None):
     file_tag = file_tags_str[file_id]
 
     if file_time is None:
-        file_path = DH.request_TM_path(file_tag)
+        file_path = DH.request_TM_path(file_tag, True)
     else:
         # Specify file_tag, latest = False and file_time
         file_path = DH.request_TM_path(file_tag, False, file_time)
--- a/flight/apps/command/commands.py
+++ b/flight/apps/command/commands.py
@@ -151,7 +151,7 @@ def REQUEST_FILE_PKT(file_id, file_time):
 
 
 def REQUEST_IMAGE():
-    raise NotImplementedError("Handled internally by comms subsystem")
+    raise NotImplementedError("Not implemented")
 
 
 def DOWNLINK_ALL(file_id, file_time=None):
--- a/flight/apps/command/commands.py
+++ b/flight/apps/command/commands.py
@@ -145,8 +145,7 @@ def REQUEST_FILE_METADATA(file_id, file_time=None):
     return [file_path]
 
 
-# TODO: Remove REQUEST_FILE_PKT REQUEST_IMAGE and DOWNLINK_ALL
-# These are handled internally in comms
+# NOTE: REQUEST_FILE_PKT REQUEST_IMAGE and DOWNLINK_ALL are handled internally in comms
 def REQUEST_FILE_PKT(file_id, file_time):
     """Requests a specific file packet from the spacecraft."""
     logger.info(f"Executing REQUEST_FILE_PKT with file_tag: {file_id}, file_tim: {file_time}")
--- a/flight/apps/command/commands.py
+++ b/flight/apps/command/commands.py
@@ -145,7 +145,7 @@ def REQUEST_FILE_METADATA(file_id, file_time=None):
     return [file_path]
 
 
-# NOTE: REQUEST_FILE_PKT REQUEST_IMAGE and DOWNLINK_ALL are handled internally in comms
+# NOTE: REQUEST_FILE_PKT handled internally in comms
 def REQUEST_FILE_PKT(file_id, file_time):
     raise NotImplementedError("Handled internally by comms subsystem")
 
@@ -154,8 +154,19 @@ def REQUEST_IMAGE():
     raise NotImplementedError("Handled internally by comms subsystem")
 
 
-def DOWNLINK_ALL():
-    raise NotImplementedError("Handled internally by comms subsystem")
+def DOWNLINK_ALL(file_id, file_time=None):
+    """Downlinks all packets for a specific file from the spacecraft."""
+    logger.info(f"Executing DOWNLINK_ALL with file_tag: {file_id} and file_time: {file_time}")
+    file_path = None
+    file_tag = file_tags_str[file_id]
+
+    if file_time is None:
+        file_path = DH.request_TM_path(file_tag)
+    else:
+        # Specify file_tag, latest = False and file_time
+        file_path = DH.request_TM_path(file_tag, False, file_time)
+
+    return [file_path]
 
 
 def get_tx_message_header():
--- a/flight/apps/command/commands.py
+++ b/flight/apps/command/commands.py
@@ -147,32 +147,15 @@ def REQUEST_FILE_METADATA(file_id, file_time=None):
 
 # NOTE: REQUEST_FILE_PKT REQUEST_IMAGE and DOWNLINK_ALL are handled internally in comms
 def REQUEST_FILE_PKT(file_id, file_time):
-    """Requests a specific file packet from the spacecraft."""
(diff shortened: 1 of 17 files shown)


flight/apps/comms/comms.py
```diff
@@ -106,13 +112,24 @@ class SATELLITE_RADIO:
...
     filepath = None
-    file_ID = 0x01
+    file_ID = 0x00
     file_size = 0
-    file_time = 1738351687
+    file_time = 0
     file_message_count = 0
 
+    """
+    NOTE: This flag can be set to force comms into DOWLINK_ALL.
```
Akshat412, line 123: This can be used for autonomously downlinking a file in orbit, completely decoupled from any GS interactions.

sa2415 (dismissed):

flight/apps/comms/comms.py
perrintong325: why is time now set to 0?

s026134 (dismissed):

flight/apps/comms/comms.py
Akshat412: file_ID, file_time etc. are file metadata parameters for the file comms is currently downlinking. When testing previously I had to hardcode these because the DH could not parse file creation time from the filename. Now that we can do that, there's no point in hardcoding / initializing these to anything.

sa2415:
Works with groundstation DOWNLINK_ALL

sa2415 (dismissed):

perrintong325 (approved):

flight/apps/comms/comms.py
```diff
@@ -400,13 +426,55 @@ def transmit_file_packet(cls):
         pkt_size = cls.file_get_packet(cls.rq_sq_cnt)
 
         # Pack header
+
+        # pkt_size + 5 is to accomodate 5 bytes of file info (file_ID, 1 byte; file_time, 4 bytes) w/ pkt_size bytes of
```
s026134, line 430: In that case I would recommend setting a macro for 5 for better readability

Ibrassow (changes requested):
. Will review tonight

s026134 (approved):

flight/apps/command/commands.py
```diff
@@ -139,38 +139,23 @@ def REQUEST_FILE_METADATA(file_id, file_time=None):
...
+    raise NotImplementedError("Handled internally by comms subsystem")
 
 
 def REQUEST_IMAGE():
-    """Requests an image from the spacecraft's internal storage."""
-    logger.info("Executing REQUEST_IMAGE")
-    # TODO: finish implementation, if we are keeping this command
-    path = DH.request_TM_path_image()
-    return [path]
+    raise NotImplementedError("Handled internally by comms subsystem")
```
Ibrassow, line 154: I disagree with this change, especially for those initial REQUEST_XXX commands. This is breaking logic already in place for no reason, breaking command logic and log paths, and this is confusing. Please revert @Akshat412 

DOWNLINK_ALL can be the exception since this is an emergency measure but there is no point for those.

flight/apps/comms/comms.py
```diff
@@ -400,13 +426,55 @@ def transmit_file_packet(cls):
...
+        # file data
         tx_header = (
             (MSG_ID.SAT_FILE_PKT).to_bytes(1, "big") + (cls.rq_sq_cnt).to_bytes(2, "big") + (pkt_size + 5).to_bytes(1, "big")
         )
 
-        # Pack entire message
+        # Pack entire message, file_array contains file info
+        cls.tx_message = tx_header + cls.file_ID.to_bytes(1, "big") + cls.file_time.to_bytes(4, "big") + cls.file_array
```
Ibrassow, line 437: Make the endianness a global constant (can be part of the HAL if necessary)

flight/apps/comms/comms.py
```diff
@@ -449,6 +517,42 @@ def check_rq_file_params(cls, packet):
...
+            logger.warning("[COMMS ERROR] Undefined TX filepath")
+            cls.dlink_all = False
+
+        else:
+            # Generate internal metadata for this filepath
+            cls.file_get_metadata()
+            cls.dlink_all = True
+            cls.int_sq_cnt = 0
+
+            logger.info(f"Starting downlink of file for {cls.file_message_count} packets")
```
Ibrassow, line 554: Even looking at this, I don't think you even need to bypass the CDH to get the file path for DOWNLINK_ALL. Cmd -> CDH gives the filepath -> Comms downlink all the packets autonomously once received from the CDH.

flight/core/data_handler.py
```diff
@@ -1341,8 +1341,9 @@ def get_closest_file_time(file_time: int, files: List[str]):
         str: file path with the closest file time to the one requested
     """
     # Search for the specific file with closest time to requested file time
+    # Do not consider the process configuration file (file[0] in file list)
     try:
-        return min(files, key=lambda f: abs(extract_time_from_filename(f) - file_time))
+        return min(files[1:], key=lambda f: abs(extract_time_from_filename(f) - file_time))
```
Ibrassow, line 1346: Please remove this change and the comment, this is breaking other logics. The function should consider only the arguments with the assuming they are all given files in the correct format (the unit tests with the same assumption). You should look at where the function is called, where the process file is already filtered to some extent: 
image

tests/test_data_handler.py
```diff
@@ -82,6 +82,7 @@ def test_extract_time_from_filename(input_filename, expected_output):
 )
 def test_get_closest_file_time(file_time, expected_output):
     files = [
+        ".data_process_configuration.json",
```
Ibrassow, line 85: see my comment above, this is not the point of the function to filter that

Ibrassow (changes requested):

perrintong325 (dismissed):

flight/core/data_handler.py
```diff
@@ -418,7 +418,7 @@ def request_TM_path(self, latest: bool = False, file_time=None) -> Optional[str]
                     transmit_file = files[1]
 
                 if file_time is not None:
-                    result_file = get_closest_file_time(file_time, files)
+                    result_file = get_closest_file_time(file_time, files[1:])
```
Ibrassow, line 421: I think the whole function - request_TM/XX_path - is stupid (not your fault lol, I'm at the origin of it) and need a refactor with a proper guarantee (right now, there is none that it won't fault). I'd suggest to hold off. I'll get a commit/PR tomorrow really quick that will solve this weirdness and guarantee a proper file

flight/core/data_handler.py
```diff
@@ -680,7 +680,7 @@ def request_TM_path(self, latest: bool = False, file_time=None) -> Optional[str]
...
```
Ibrassow, line 683: ^Same as previous comment

flight/core/data_handler.py
Akshat412: That sounds good, but do note that for comms at least there is error checking to ensure a filename is not None or something. As long as get_closest_file_time doesn't return a bogus filename we should be good.

Akshat412: Where is the PR :eyes:

perrintong325 (approved):

Ibrassow (approved):
Small comment change to do then it's good to go. Do you have results of the DOWNLINK_ALL RX on the GS side to post on this PR (list of files downlinked + screenshot of one you decoded)

sa2415 (dismissed):

s026134 (approved):

Akshat412:
image

PR #275, opened 2025-04-20
Burnwire sequence

perrintong325:
fixed burn wire driver that doesn't like when you only write 1 bit. tested on burnwire driver with fish wires. Added burn wire sequence to command

### DIFF
--- a/flight/tasks/command.py
+++ b/flight/tasks/command.py
@@ -59,10 +59,13 @@ class Task(TemplateTask):
         # DEPLOYMENT SEQUENCE
         # ------------------------------------------------------------------------------------------------------------------------------------
         burn_wires = SATELLITE.BURN_WIRES
-        if self.deploymentPWM != 0:
-            # Disable previous PWM to prevent brown out
+        if self.deploymentPWM == 0:
+            # Enable the first PWM
+            burn_wires.set_pwm(self.deploymentPWM, _BURN_WIRE_STRENGTH)
+        elif self.deploymentPWM < _PWM_MAX:
+            # Disable previous PWM and enable current one
             burn_wires.set_pwm(self.deploymentPWM - 1, 0)
-        burn_wires.set_pwm(self.deploymentPWM, _BURN_WIRE_STRENGTH)
+            burn_wires.set_pwm(self.deploymentPWM, _BURN_WIRE_STRENGTH)
         burn_wires.enable_driver()
         self.deploymentPWM += 1  # Increment PWM for next deployment
 
--- a/flight/tasks/command.py
+++ b/flight/tasks/command.py
@@ -21,6 +21,7 @@ _TPM_INIT_TIMEOUT = const(10)  # seconds
 _EXIT_STARTUP_TIMEOUT = const(5)  # seconds
 _BURN_WIRE_STRENGTH = const(7)  # 0-255
 _DEPLOYMENT_INTERVAL = const(5)  # seconds
+_PWM_MAX = const(3)  # Maximum PWM value for deployment
 
 
 class Task(TemplateTask):
@@ -138,11 +139,11 @@ class Task(TemplateTask):
 
                 if SATELLITE.BURN_WIRES_AVAILABLE:  # TODO: add deployment flag
                     # Deployment finished when the deployment PWM reaches 3
-                    if self.deploymentPWM == 3 and deployment_time_check:
+                    if self.deploymentPWM == _PWM_MAX and deployment_time_check:
                         self.log_info("Deployment complete")
                         self.deployment_done = True
                         SATELLITE.BURN_WIRES.disable_driver()
-                    elif self.deploymentPWM < 3 and deployment_time_check:
+                    elif self.deploymentPWM < _PWM_MAX and deployment_time_check:
                         self.log_info(f"Deployment sequence: {self.deploymentPWM}")
                         self.last_deployment_time = TPM.monotonic()
                         self.deployment_sequence()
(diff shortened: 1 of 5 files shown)


flight/tasks/command.py
```diff
@@ -47,6 +48,21 @@ def __init__(self, id):
...
+        # ------------------------------------------------------------------------------------------------------------------------------------
+        if SATELLITE.BURN_WIRES_AVAILABLE:
+            self.log_info("Burn wires available, deploying...")
+            burn_wires = SATELLITE.BURN_WIRES
+            burn_wires.set_pwm(0, _BURN_WIRE_STRENGTH)
+            burn_wires.set_pwm(1, _BURN_WIRE_STRENGTH)
+            burn_wires.set_pwm(2, _BURN_WIRE_STRENGTH)
+            burn_wires.enable_driver()
+            TPM.sleep(5)
+            burn_wires.disable_driver()
```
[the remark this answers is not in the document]
perrintong325: r u on crack wt r u on bro

flight/tasks/command.py
```diff
@@ -47,6 +48,25 @@ def __init__(self, id):
...
+            burn_wires.set_pwm(0, _BURN_WIRE_STRENGTH)
+            burn_wires.enable_driver()
+            TPM.sleep(5)
+            burn_wires.set_pwm(0, 0)
+            burn_wires.set_pwm(1, _BURN_WIRE_STRENGTH)
+            TPM.sleep(5)
+            burn_wires.set_pwm(1, 0)
+            burn_wires.set_pwm(2, _BURN_WIRE_STRENGTH)
+            TPM.sleep(5)
+            burn_wires.disable_driver()
```
Ibrassow, line 67: No sleep please. Use time-based checks across task run cycles to enable/disable  the burnwires and set the pwm

Ibrassow (changes requested):

flight/tasks/command.py
Akshat412: @perrintong325 when the WDT is active, this will make it trigger. It's best to put all this in a function and implement it as a state machine where the state changes across task runs based on the clock time.

flight/tasks/command.py
```diff
@@ -111,9 +129,31 @@ def startup(self):
...
+                # check if the deployment is ready to be performed
+                deployment_time_check = (
+                    (TPM.monotonic() - self.last_deployment_time) >= _DEPLOYMENT_INTERVAL
+                    if self.last_deployment_time is not None
+                    else True
+                )
+
+                if SATELLITE.BURN_WIRES_AVAILABLE:  # TODO: add deployment flag
+                    # Deployment finished when the deployment PWM reaches 3
+                    if self.deploymentPWM == 3 and deployment_time_check:
```
Ibrassow, line 141: make the "3" a global variable or class attribute to be tuned --> `NUM_DEPLOYMENT_PWM = 3`

flight/tasks/command.py
```diff
@@ -111,9 +129,31 @@ def startup(self):
...
+                        self.log_info("Deployment complete")
+                        self.deployment_done = True
+                        SATELLITE.BURN_WIRES.disable_driver()
+                    elif self.deploymentPWM < 3 and deployment_time_check:
```
Ibrassow, line 145: make the "3" a global variable or class attribute to change it easily --> `NUM_TRIAL_DEPLOYMENT_PWM = 3`

flight/tasks/command.py
```diff
@@ -111,9 +129,31 @@ def startup(self):
...
+                        self.log_info(f"Deployment sequence: {self.deploymentPWM}")
+                        self.last_deployment_time = TPM.monotonic()
+                        self.deployment_sequence()
+                else:
+                    self.log_info("Skipping deployment sequence...")
```
Ibrassow, line 150: `self.log_info("Skipping deployment sequence... Burn wires not available")`

flight/tasks/command.py
```diff
@@ -19,6 +19,8 @@
 
 _TPM_INIT_TIMEOUT = const(10)  # seconds
 _EXIT_STARTUP_TIMEOUT = const(5)  # seconds
+_BURN_WIRE_STRENGTH = const(7)  # 0-255
+_DEPLOYMENT_INTERVAL = const(5)  # seconds
```
Ibrassow, line 23: is that enough? we can make the STARTUP state last longer if needed too

flight/tasks/command.py
perrintong325: it's not necessarily true because we need to add deployment flags(ie if we reboot the sat we can skip this). cc @Akshat412

flight/tasks/command.py
perrintong325: it burnt it when we tried. 7 was the highest I can get without browning out. might hv to make the intervals longer still need to try this again.

flight/tasks/command.py
Ibrassow: I'd still change the message if the deployment flags are missing (in this case, you'd also have another log message). We shouldn't be logging (or writing any software, for that matter) based on features that don't exist yet

kmkar31 (approved):

flight/tasks/command.py
Akshat412: I can add this in within this PR, it's not a big change.

Akshat412: But also agree with the above. Deployment flag is a very different case from the burn wires not being available (which can legitimately happen in dev, this is an I2C device). Both situations need very different log messages.

Akshat412 (approved):

Ibrassow (approved):

kmkar31 (approved):

PR #276, opened 2025-04-22
ADCS bug fixes 5

kmkar31:
Fixes faults in 
- ACS
- MEKF


kmkar31:
Detumbling Fix:
- Signs were wrong on coil allocations i.e., if +2.5V is provided to XP and XM coils, fields cancel out. Fixed this by assigning opposite signed voltages to coils on the same axis
- spin stabilization gain was tuned to reduce detumbling time. Now detumbles in the range of 40 min to 2 hours

Akshat412:
I think you should merge from main......

kmkar31:
All of ACS work
control_input
ing
omega
sun_point

PR #280, opened 2025-04-26
HAL Monitor

perrintong325:
- HAL Monitor
  - monitors all device errors by request from HAL
  - Logs all device errors, error counts, dead status and peripheral reboot count to sd card
  - graceful reboot when power cycling peripheral line
  - decides how each error should be proceeded
  - regular 24hr reboots

 - All drivers
   - device_errors fn to sample error register

- cubesat
  - new fn to update each device
  - graceful reboot fn for hal monitor
 
- errors
  - all errors from drivers (MIDDLEWARE IS NOW COMPLETELY GONE)
 
- Data Handler
  - fn to support graceful reboot
 
tested on argus 3 with no fuel gauge connected or faulty radio. regular reboot tested with 1 min intervals.


flight/tasks/hal_monitor.py
```diff
@@ -0,0 +1,133 @@
...
+
+    ######################## HELPER FUNCTIONS ########################
+
+    def idx_to_hal_name(self, idx: int):
+        """Return the HAL_IDX field name for a given index, or None if not found."""
+        return _HAL_IDX_INV.get(idx)
+
+    def close_data_process(self):
+        for data_process in DH.get_all_data_processes():
+            data_process.close()
```
Ibrassow, line 36: this is not needed, remove

flight/core/data_handler.py
```diff
@@ -1015,6 +1015,28 @@ def get_data_process(cls, tag_name: str) -> DataProcess:
...
+    @classmethod
+    def restore_data_process_files(cls) -> None:
+        """
+        Restores the data process files by reinitializing them.
+
+        Returns:
+            None
+        """
+        for tag_name in cls.data_process_registry:
+            cls.data_process_registry[tag_name].try_to_reuse_latest_file()
```
Ibrassow, line 1038: remove, this is not needed. All data processes already do it at init time

flight/core/data_handler.py
```diff
@@ -1015,6 +1015,28 @@ def get_data_process(cls, tag_name: str) -> DataProcess:
         """
         return cls.data_process_registry[tag_name]
 
+    @classmethod
+    def graceful_shutdown(cls) -> None:
+        """
+        Gracefully shuts down all data processes by closing their files.
```
Ibrassow, line 1021: 'WARNING: should not be used unless for self-driven reboot. Gracefully shuts down all data processes by closing their files."

emulator/cubesat.py
```diff
@@ -261,32 +306,55 @@ def BATTERY_HEATERS(self):
...
+    def SAMPLE_DEVICE_ERRORS(self) -> dict[str, list[int]]:
+        """SAMPLE_DEVICE_ERRORS: Sample the device errors"""
+        errors = {}
+        for name, device in self._device_list.items():
+            if device.ASIL != ASIL0:
+                if device.device is None and device.dead is False and device.error == Errors.DEVICE_NOT_INITIALISED:
+                    errors[name] = [Errors.DEVICE_NOT_INITIALISED]
+                elif device.device is not None and device.dead is False:
+                    errors[name] = device.device.device_errors
+        return errors
```
Ibrassow, line 360: this is making lots of dynamic allocations every time this is called. Can you at least pre-construct the dict with the keys and keep it stored with the class? Then you just fill the array. (Another level is to pre-allocate even the array tbh and just fill the elements of interest and have a sentinel value)

flight/apps/telemetry/constants.py
```diff
@@ -153,6 +140,110 @@ class STORAGE_IDX:
     DIR_SIZE = const(1)
 
 
+class HAL_IDX:
```
Ibrassow, line 143: Can you have a special error code specifically for DEAD so you can save 1/3 of the array? ~~ 33 bytes * log time * nb files etc That's a lot of space wasted for nothing (RAM, global space, and SD) Once DEAD, you don't care about the other errors (well, you won't have them)

flight/tasks/hal_monitor.py
```diff
@@ -0,0 +1,133 @@
...
+        if result == Errors.NO_REBOOT:
+            self.log_info(f"Device {device_name} has {device_error}, no reboot occured")
+            SATELLITE.update_device_error(device_name, device_error)
+        elif result == Errors.REBOOT_DEVICE:
+            self.log_info(f"Rebooted {device_name} due to error {device_error}")
+        elif result == Errors.GRACEFUL_REBOOT:
+            self.peripheral_reboot_count += 1
+            DH.graceful_shutdown()
+            SATELLITE.graceful_reboot_devices(device_name)
+            DH.restore_data_process_files()
```
Ibrassow, line 85: DH stuff not needed. Can you explain the context here? DH will never need a "reboot" for any peripheral reboot. Only that might need a DH handling is the SD card, but that's it. For the SD card case, I can handle it properly in a separate PR (you can raise an issue assigned to me) as this does not involve only close/opening the files

flight/tasks/hal_monitor.py
```diff
@@ -0,0 +1,133 @@
+# HAL Monitor Task
+# This task is responsible for monitoring the health of the hardware abstraction layer (HAL),
+# performing diagnostics in case of failure, and reporting/logging HAL status.
```
Ibrassow, line 3: Can you add detailed docstrings and comments throughout the task to parse a bit more easily. For such a critical thing, I want to make sure we cover everything with no assumption

Ibrassow (changes requested):

flight/tasks/hal_monitor.py
perrintong325: we need this every time we reboot the peripheral line because that reboots the sd card. if we dont have this everything shits itself

flight/core/data_handler.py
perrintong325: we need this every time we reboot the peripheral line because that reboots the sd card. if we dont have this everything shits itself

flight/apps/telemetry/constants.py
perrintong325: no because we want to know what caused it to die. Akshat and I talked about this and think we should keep it as is.

flight/tasks/hal_monitor.py
perrintong325: well I can't remove this for this pr because this stops everything from running properly. you can change it in another pr but getting rid of this would break the entire pr.

flight/core/task_configuration.py
```diff
@@ -17,4 +18,5 @@
     TASK.ADCS: {"Task": adcs, "Frequency": 5, "Priority": 1},
     TASK.GPS: {"Task": gps, "Frequency": 0.025, "Priority": 3, "ScheduleLater": True},
     TASK.PAYLOAD: {"Task": payload, "Frequency": 0.1, "Priority": 3, "ScheduleLater": True},
+    TASK.HAL_MONITOR: {"Task": hal_monitor, "Frequency": 0.5, "Priority": 1},
```
kmkar31, line 21: Is this feasible and why do we want it this fast? I'd assume a slower hal monitor task will only delay reboots and error handling which seems fine. I'd say make it once a minute

flight/hal/drivers/sx126x.py
```diff
@@ -1480,6 +1483,7 @@ def begin(
         ASSERT(state)
 
         state = self.setOutputPower(power)
+        print(state)
```
kmkar31, line 1486: Does this print statement have to go?

flight/core/task_configuration.py
perrintong325: we talked about it and we said 2s yesterday and it runs fine because it's blcoking. we talked again and think we'll stick with 10s

flight/apps/telemetry/constants.py
Ibrassow: You will get that last error in your telemetry and the file you downlinked. Once your device is dead, you're still wasting all that memory for info you already have

emulator/drivers/imu.py
```diff
@@ -1,4 +1,13 @@
+from hal.drivers.errors import Errors
 from numpy import array
+from ulab import numpy as np
+
+# Failure Probabilities:
+_prob_ = np.array([2, 2, 0.1])  # % of devices that throw [error_code, drop_cmd, fatal_err] fault in a day
+
+_scale_ = -86400 / (np.log(1 - (0.01 * _prob_)))  # exponential distribution scale
+
```
Ibrassow, line 9: Have this in a separate config file (in your format of choice) for all devices for the SIL to be read at SIL init. We shouldn't have to change the SW for those parameters

flight/apps/telemetry/constants.py
Ibrassow: thanks

flight/hal/argus_v3.py
```diff
@@ -716,9 +780,20 @@ class ArgusV3Error:
     }
 
     ASIL_THRESHOLDS = {
-        ASIL1: const(5),  # ASIL 1: Reboot after 5 errors
-        ASIL2: const(3),  # ASIL 2: Reboot after 3 errors
-        ASIL3: const(2),  # ASIL 3: Reboot after 2 errors
+        ASIL1: const(15),  # ASIL 1: Reboot after 15 errors
+        ASIL2: const(9),  # ASIL 2: Reboot after 9 errors
+        ASIL3: const(6),  # ASIL 3: Reboot after 6 errors
```
Ibrassow, line 785: what was the reasoning behind those values?

flight/hal/drivers/bmx160.py
```diff
@@ -666,6 +667,18 @@ def generic_setter(
         else:
             settingswarning(warning_interp)
     """
+    ######################## ERROR HANDLING ########################
+
+    @property
+    def device_errors(self):
+        results = []
+        if self.error_code != BMX160_OK:
+            results.append(Errors.IMU_ERROR_CODE)
```
Ibrassow, line 676: `Errors.IMU_ERROR_CODE` what does that mean? Can we use a more descriptive name?

flight/hal/drivers/drv8235.py
```diff
@@ -37,12 +79,12 @@
...
-_OCP = const(0x03)
-_OVP = const(0x04)
-_TSD = const(0x05)
-_NPOR = const(0x06)
+# _FAULT = const(0x01)
+# _STALL = const(0x02)
+# _OCP = const(0x03)
+# _OVP = const(0x04)
+# _TSD = const(0x05)
+# _NPOR = const(0x06)
```
Ibrassow, line 87: remove dead code/comments

flight/hal/drivers/errors.py
```diff
@@ -8,37 +8,64 @@ class Errors:
...
+    TORQUE_COIL_THERMAL_SHUTDOWN = const(0xB)
+    TORQUE_COIL_EXTENDED_CURRENT_LIMIT_EVENT = const(0xC)
+    TORQUE_COIL_THROTTLE_OUTSIDE_RANGE = const(0xD)
+    TORQUE_COIL_STALL_EVENT = const(0xE)
+
+    # Light Sensor Errors
+    LIGHT_SENSOR_HIGHER_THAN_THRESHOLD = const(0x8)
+    LIGHT_SENSOR_LOWER_THAN_THRESHOLD = const(0x9)
+    LIGHT_SENSOR_OVERFLOW = const(0xA)
+    LIGHT_SENSOR_COVERSION_READY = const(0xB)
```
Ibrassow, line 56: This is an odd name, it doesn't sound like an error (+ typo I think)

flight/hal/drivers/errors.py
```diff
@@ -8,37 +8,64 @@ class Errors:
...
 
     # RTC Errors
-    RTC_LOST_POWER = const(0x10)  # placeholder value
-    RTC_BATTERY_LOW = const(0x11)  # placeholder value
+    RTC_LOST_POWER = const(0x8)
+    RTC_BATTERY_LOW = const(0x9)
 
     # GPS Errors
-    GPS_UPDATE_CHECK_FAILED = const(0x20)  # placeholder value
+    GPS_UPDATE_CHECK_FAILED = const(0x8)
```
Ibrassow, line 29: suspicious that there is only 1 error for the GPS...

flight/tasks/hal_monitor.py
```diff
@@ -0,0 +1,183 @@
...
+                DH.restore_data_process_files()
+                self.log_info("Gracefully rebooted peripheral power line.")
+            self.graceful_reboot_counter = 0
+        else:
+            self.graceful_reboot_counter += 1
+
+        self.log_device_status()
+        DH.log_data(self.log_name, self.log_data)
+        # regular reboot every 24 hours
+        # TODO: delay this if the satellite is at a ground pass
```
Ibrassow, line 179: Raise an issue for  this and reference it as a reply to this comment please

Ibrassow (changes requested):
Please, include hardware fault injection results before any approval

perrintong325:
[hal_monitor_2hr_vibe.txt](https://github.com/user-attachments/files/20071834/hal_monitor_2hr_vibe.txt)
2 hour ish day in the life test on vibe build
bugs:
- using same constant value causing issues when there's the same value for something else, fixed
- rebooting causes some peripherals to be not immediately accessible, to be fixed
notes:
- "Turned on TORQUE_ZP" really just means turned on all coil, will update message
- graceful reboot periph worked as expected

flight/hal/drivers/errors.py
perrintong325: the thing is these are not really errors but flags. i only put them together because they are raised from the same place and i only put it there if other subsystems want to use them. they can be removed

Ibrassow: In general, we should avoid adding stuff that has no immediate use

perrintong325:
[hal_monitor_remove_zp.txt](https://github.com/user-attachments/files/20074335/hal_monitor_remove_zp.txt)
hardware injection test, removed zp halfway through, hal monitor successful caught and disabled the devices on the board, other tasks didn't shit itself
image

flight/hal/argus_v3.py
perrintong325: i have to decrease them back down to what it was, but really just how many times it's really just out of our ass on what we thought is right

perrintong325:
image
manual object wrapper test as imu didn't die in day in the life test by always raising an error in the driver(mag fn):
successfully caught and logged driver level errors, debug messages to show it was correctly logged

perrintong325:
https://drive.google.com/drive/folders/128IwbFlI7A-oHIl0U3m1HFzCDHd03HMd?usp=drive_link
15 hour log, nothing special showed up

emulator/drivers/imu.py
perrintong325: fixed

emulator/drivers/failure_prob.py
```diff
@@ -0,0 +1,29 @@
...
+    _prob_ = 2 * np.ones((2,))  # % of devices that throw [lost_power, battery_low] fault in a day
+
+    scale = -86400 / (np.log(1 - (0.01 * _prob_)))  # exponential distribution scale
+
+
+class imu_prob:
+    # Failure Probabilities:
+    _prob_ = np.array([2, 0.1])  # % of devices that throw [drop_cmd, fatal_err] fault in a day
+
+    scale = -86400 / (np.log(1 - (0.01 * _prob_)))  # exponential distribution scale
```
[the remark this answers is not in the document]
Ibrassow: ALSO, this file reading should be done once at the emulator class level, and then pass the scale thing through arguments in their constructor. No need for unnecessary imports at every emulated drivers

Ibrassow (commented):
Mmh big PR are really bad in general for the reviewers tbh. Let's try to make small incremental PRs next time. 

For all the issues you might have highlighted, can you please raise issues for them for further investigation? Even for those that didn't show up again, better be safe than sorry

Ibrassow:
@ChromeUniverse Can you take over and address the comments?

PR #281, opened 2025-04-27
added commit hash and branch to build and boot

perrintong325:
it still builds

### DIFF
--- a/build_tools/build-emulator.py
+++ b/build_tools/build-emulator.py
@@ -8,6 +8,24 @@ def check_directory_location(source_folder):
         raise FileNotFoundError(f"Source folder {source_folder} not found")
 
 
+def get_commit_hash():
+    try:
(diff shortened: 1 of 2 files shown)


build_tools/build-emulator.py
```diff
@@ -8,6 +8,24 @@ def check_directory_location(source_folder):
         raise FileNotFoundError(f"Source folder {source_folder} not found")
 
 
+def get_commit_hash():
+    try:
+        commit_hash = os.popen("git rev-parse --short HEAD").read().strip()
+        return commit_hash
```
Ibrassow, line 14: can you display instead the full commit hash instead of the shortened version?

Ibrassow (changes requested):

Ibrassow (approved):

Akshat412 (approved):

PR #327, opened 2025-11-09
ADCS:  deployment adjustment

prchall123:
This branch aims to achieve the following:

- [x] Sim update to model scenarios with panel deployment failures and deployment sensors
- [x] Updating emulator in fsw repo to add deployment sensor
- [x] ADCS adjusting inertia matrix based on deployment sensor readings
- [x] Demonstrating ADCS response through MC simulation with single panel deployment failure added to SIL campaign set
- [ ] Hardware validation of ADCS response to deployment failure

Simulation with 50 MC runs and failed deployment of xp panel.
image


prchall123:
Deployment is added to the sil in the controller mode branch. Why the adjustment to deployment failure won't be included is also discussed there. If the panels don't deploy, its best to fall back to a detumbling controller than try the switching controller

PR #341, opened 2025-11-22
Data Handler updated file process and Comms access updates

perrintong325:
Data Handler
- Removed image process, generic file proccess
- internal file container, optimized for comms packetisation
  - file process' log fn will always write a .bin file, identified using a file header("ARGUS" currently, might be changed to something more unique avoid to accidental mishandling)
  - data is stored as packets for comms purposes, maxing out at 240 bytes per payload (ie each log can only store 240 bytes)
  - Each packet is internally packed to 240 bytes if payload is smaller than 240 bytes. This is to accomplish O(1) access time. Prioritizing compute over storage.
  - This file structure is local to DH. Get_packet fn will output a tuple, which includes the payload length and payload(unpadded). This is also designed for comms optimization.
- DH now supports non-jpg files. FileProcess can read output files not written by DH and non-jpg files. Get_packet fn will simply read 240 bytes and output it as such.

Comms
- Updated to use new DH structure
- Tested with GS

### DIFF
--- a/flight/apps/comms/comms.py
+++ b/flight/apps/comms/comms.py
@@ -12,6 +12,7 @@ from apps.command.constants import file_ids_str
 from core import logger
 from core.data_handler import DataHandler as DH
 from core.data_handler import extract_time_from_filename
+from core.satellite_config import comms_config as CONFIG
 from hal.configuration import SATELLITE
 from micropython import const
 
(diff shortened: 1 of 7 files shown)


flight/apps/comms/comms.py
```diff
@@ -69,7 +70,7 @@ class MSG_ID:
     """
 
     # Header ID for Argus - THIS MUST BE UNIQUE FOR EACH SPACECRAFT
-    ARGUS_ID = 0x01
+    ARGUS_ID = 0x00
```
prchall123: we might want to put this in the configuration/flight.yaml to keep in mind

Mahii11 (approved):

PR #362, opened 2026-02-15
Controller mode

### LINKED ISSUE #326: ADCS: IMU issues
The following issues were detected while running the fsw on the assembly:

- IMU on assembly board is outputting very high values at rest (~ 30 deg/s). 
- The parameters in the bmx160 driver do not seem to be well configured: the range and the sensitivity do not appear to match with the number of bits of the gyro. Sensitivity is set to RFS125 (131.2), scale is set to RFS500.

In [bmx160.py](https://github.com/cmu-argus-2/FSW-mainboard/tree/main/flight/hal/drivers/bmx160.py), lines 454-463:

```python
def init_gyro(self):
        # BW doesn't have an interface yet
        self._gyro_bwmode = _BMX160_GYRO_BW_NORMAL_MODE
        # These rely on the setters to do their magic.
        self.gyro_range = _BMX160_GYRO_RANGE_500_DPS
        # self.GYR_SCALAR = 1
        # self.GYR_SCALAR = 1/GyroSensitivity2DegPerSec_values[1]
        self.gyro_odr = 25
        self.gyro_powermode = _BMX160_GYRO_NORMAL_MODE
```

In lines 273-284 and 305-324: 

```python
GyroSensitivity2DegPerSec_values = [
    16.4,
    32.8,
    65.6,
    131.2,
    262.4,
]  # Section 1.2, Table 3
class BMX160:
    """
    Driver for the BMX160 accelerometer, magnetometer, gyroscope.

    In the buffer, bytes are allocated as follows:
        - mag 0-5
        - rhall 6-7 (not relevant?)
        - gyro 8-13
        - accel 14-19
        - sensor time 20-22
    """

    # multiplicative constants

    # NOTE THESE FIRST TWO GET SET IN THE INIT SEQUENCE
    DEG_TO_RAD = pi / 180
    GYR_SCALAR = (1 / GyroSensitivity2DegPerSec_values[4]) * DEG_TO_RAD  # radians/s
```
From the bmx160 [datasheet](https://www.mouser.com/pdfdocs/BST-BMX160-DS000-11.pdf?srsltid=AfmBOorUUhc46hwK4kUDEvK7akZW-TXwvU6X61NDy84Xr05XneEPwfKO):

Image

- (optional) if possible, it would be useful to change the sensitivity of the imu depending on the global mode (detumbling = higher range/less resolution at 2000dps, nominal lower range/higher resolution). If not, it's preferable to configure the gyro to the highest range to be conservative.

### LINKED ISSUE #358: ADCS: Default to B-dot/B-cross
Data collection from the IMU on the mainboard and Jetson showed a substantial bias in both the magnetometer and gyro, that may not allow the spin-stabilized pointing controller to work effectively below 3 deg/s. 

In detumbling mode, the controller should always be b-dot.

In nominal, a flag should exist in ADCS to select the controller between a) b-dot b) b-cross and c) lyapunov spin-stabilized sun-pointing.

B-dot will be more robust to imu bias, though sensitive to magnetometer noise and won't drive the system to a lower rate than that of the magnetic field's rate of change times the gain.
B-cross will be more sensitive to the imu bias and to getting stuck tracking the uncontrollable subspace, but should drive the system to an angular rate on the order of the gyro bias. Which one will work best at low angular rates (<3 deg/s) is to be determined.
The spin-stabilized sun-pointing mode should be set at a higher target angular rate to work with the worst case bias of the bmx160.

A ground command can be set to choose the nominal controller mode.

prchall123:
Closes #358 , Closes #326 , Closes #327 

# Pull Request for Controller Mode branch

This pull request will add the following features:

1. A controller mode flag to switch between b-dot, b-cross or the lyapunov switching controller. B-dot will be set as the default controller. Code was written to keep the controller mode change persistent across reboots, which writes the controller mode to a 1-byte file in sd/config. It is kept in a separate folder that is not a registered data process, it logs an error at the start due to there not being a config json for it, but other than that it does not disturb the rest of the FSW.
2. SIL integration of the deployment sensors, with the possibility of simulating a deployment failure
3. Fix of the magnetorquer/magnetometer switching function to a more reliable approach
4. Adjustment of the IMU scale settings according to the mode (if detumbling, the scale is maxed, if nominal, resolution is maxed)
5. SIL integration of the RTC

A feature that was implemented in this branch but has been removed is the adjustment of the inertia matrix based on the output of the deployment sensors. The only controller affected by this is the lyapunov switching controller, which relies on this information and may not work accurately if the panels fail to deploy. If the panels fail to deploy, however, it may not be worth employing the controller anyway, both because the +z direction wont be producing that much more power than the others, and because the axis of inertia change alignment.

# Controller Modes

Here below are the results of the different controller modes.

## B-dot Monte Carlo

Below are the results for the b-dot algorithm. It only detumbles the spacecraft to less than 4 degrees per second because the gyro bias of the bmx160 is between +-3 deg/s, and this offers some margin.

image

## B-cross Monte Carlo

Below are the results for the b-cross algorithm. 
image

## Lyapunov Switching Controller

The results for the lyapunov switching controller are shown below. The target angular velocity was raised to 20 degrees per second to account for the gyro bias. The controller takes longer than before to converge to the target equilibrium with the conservatively high gyro bias. A few of the simulations still hadn't reached the target sun pointing after 10 hours, though they were getting close to it.

image

# Magnetorquer/magnetometer switching

The approach for the magnetometer/torquer switching was previously based on a counter system - with the adcs running at 5 hz, the adcs task would count 3 runs and switch the coils off, and once it got to 5 it would run the magnetometer and then restart. The rtc has a second resolution and cannot be used to time the coils and imu. A function was created using the microcontroller clock for a ms resolution that is used to time the switching. Now, the coils run until 0.8s have passed since they were turned on, after which they are switched off. Once 0.2 s have passed since they were switched off, a magnetometer measurement is taken and then the coils are allowed to run again. The SIL is considering an exagerated effect of the coils on the magnetometer to check that it only samples when the coils are off.

image

# Controller Mode switching by command

A command has been added that allows for switching the controller mode between b-dot, b-cross and the lyapunov sun pointing controller. As exemplified below, we send a command through the GSW frontend:

image

The result is the SC receiving the command and changing the controller mode, as seen in this log:

```
[2000-01-01 00:00:26][INFO] [6][ADCS] Logging ADCS data at time 946684826
[2000-01-01 00:00:26][INFO] [6][ADCS] ADCS Mode : 3
[2000-01-01 00:00:26][INFO] [6][ADCS] Controller Mode : 0
[2000-01-01 00:00:26][INFO] [6][ADCS] Gyro Ang Vel : [0.00639158, -0.00239684, -0.00106526]
[2000-01-01 00:00:26][INFO] [6][ADCS] Mag Field : [7.00625e-05, -2.49375e-05, 1.9375e-06]
[2000-01-01 00:00:26][INFO] [6][ADCS] Sun Vector : [0.0, 0.0, 0.0]
[2000-01-01 00:00:26][INFO] [6][ADCS] Sun Status : 51
[2000-01-01 00:00:26][INFO] [6][ADCS] Gyro Status : 0
[2000-01-01 00:00:26][INFO] [6][ADCS] Mag Status : 0
[2000-01-01 00:00:26][INFO] [0][COMMAND] Time: 946684826
[2000-01-01 00:00:26][INFO] [0][COMMAND] Time since boot: 27
[2000-01-01 00:00:26][INFO] [0][COMMAND] GLOBAL STATE: NOMINAL.
[2000-01-01 00:00:26][INFO] [0][COMMAND] RAM USAGE: 80%
[2000-01-01 00:00:26][INFO] [4][COMMS] Checking transmit queue for packets to send...
[2000-01-01 00:00:26][INFO] [4][COMMS]   Transmit queue size: 0
[2000-01-01 00:00:26][INFO] [4][COMMS] Checking for incoming messages from GS...
[2000-01-01 00:00:27][INFO] [COMMS] Command authentication passed
[2000-01-01 00:00:27][INFO] Received callsign: CSXXXX
[2000-01-01 00:00:27][INFO] Received raw packet: b'CSXXXX@\x1f\x01'
[2000-01-01 00:00:27][INFO] Unpacked message object: Command('ADCS_CTRL_MODE', id=31, args: ['mode_id'])
[2000-01-01 00:00:27][INFO] [4][COMMS] Received command from GS: Command('ADCS_CTRL_MODE', id=31, args: ['mode_id'])
[2000-01-01 00:00:27][CRITICAL] Error: Data process comms not registered!
[2000-01-01 00:00:27][INFO] [9][PAYLOAD] Payload state: OFF
[2000-01-01 00:00:27][INFO] [6][ADCS] Logging ADCS data at time 946684827
[2000-01-01 00:00:27][INFO] [6][ADCS] ADCS Mode : 3
[2000-01-01 00:00:27][INFO] [6][ADCS] Controller Mode : 0
[2000-01-01 00:00:27][INFO] [6][ADCS] Gyro Ang Vel : [0.0066579, -0.00266316, -0.000798948]
[2000-01-01 00:00:27][INFO] [6][ADCS] Mag Field : [7.80625e-05, -1.79375e-05, 2.3125e-06]
[2000-01-01 00:00:27][INFO] [6][ADCS] Sun Vector : [0.0, 0.0, 0.0]
[2000-01-01 00:00:27][INFO] [6][ADCS] Sun Status : 51
[2000-01-01 00:00:27][INFO] [6][ADCS] Gyro Status : 0
[2000-01-01 00:00:27][INFO] [6][ADCS] Mag Status : 0
[2000-01-01 00:00:27][INFO] [4][COMMS] Checking transmit queue for packets to send...
[2000-01-01 00:00:27][INFO] [4][COMMS]   Transmit queue size: 0
[2000-01-01 00:00:27][INFO] [4][COMMS] Checking for incoming messages from GS...
[2000-01-01 00:00:28][INFO] [0][COMMAND] Popped command from queue: Command('ADCS_CTRL_MODE', id=31, args: ['mode_id']) with status: 0
[2000-01-01 00:00:28][INFO] [0][COMMAND]   Arguments: {'mode_id': 1}
[2000-01-01 00:00:28][INFO] Processing command: ADCS_CTRL_MODE with arguments: [1]
[2000-01-01 00:00:28][INFO] and precondition: valid_adcs_mode
[2000-01-01 00:00:28][INFO] Executing command function: ADCS_CTRL_MODE
[2000-01-01 00:00:28][INFO] Executing ADCS_CTRL_MODE with mode_id: 1
[2000-01-01 00:00:28][INFO] Added ack obj to transmit queue: Ack('rid=0', args=[31, 1, True])
[2000-01-01 00:00:28][INFO] Command execution successful
[2000-01-01 00:00:28][INFO] [9][PAYLOAD] Payload state: OFF
[2000-01-01 00:00:28][INFO] [3][OBDH] Data processes: ['payload_tm', 'payload_od', 'eps_warning', 'payload_requests', 'hal', 'cdh', 'eps', 'cmd_logs', 'adcs', 'img']
[2000-01-01 00:00:28][INFO] [6][ADCS] Logging ADCS data at time 946684828
[2000-01-01 00:00:28][INFO] [6][ADCS] ADCS Mode : 3
[2000-01-01 00:00:28][INFO] [6][ADCS] Controller Mode : 1
[2000-01-01 00:00:28][INFO] [6][ADCS] Gyro Ang Vel : [0.0066579, -0.00292947, -0.000798948]
[2000-01-01 00:00:28][INFO] [6][ADCS] Mag Field : [7.05625e-05, -2.19375e-05, 2.6875e-06]
[2000-01-01 00:00:28][INFO] [6][ADCS] Sun Vector : [0.0, 0.0, 0.0]
[2000-01-01 00:00:28][INFO] [6][ADCS] Sun Status : 51
[2000-01-01 00:00:28][INFO] [6][ADCS] Gyro Status : 0
[2000-01-01 00:00:28][INFO] [6][ADCS] Mag Status : 0
```

The SC is able to detect invalid controller mode switch commands. The GS won't be able to command non-positive integers above 255. With controller mode ids between 0 and 2, if any mode is sent with id greater than 2, the SC will not accept it:

```
[2000-01-01 00:00:11][INFO] [6][ADCS] Logging ADCS data at time 946684811
[2000-01-01 00:00:11][INFO] [6][ADCS] ADCS Mode : 3
[2000-01-01 00:00:11][INFO] [6][ADCS] Controller Mode : 0
[2000-01-01 00:00:11][INFO] [6][ADCS] Gyro Ang Vel : [0.00612526, -0.00239684, -0.00133158]
[2000-01-01 00:00:11][INFO] [6][ADCS] Mag Field : [7.30625e-05, -1.59375e-05, 2.6875e-06]
[2000-01-01 00:00:11][INFO] [6][ADCS] Sun Vector : [0.0, 0.0, 0.0]
[2000-01-01 00:00:11][INFO] [6][ADCS] Sun Status : 51
[2000-01-01 00:00:11][INFO] [6][ADCS] Gyro Status : 0
[2000-01-01 00:00:11][INFO] [6][ADCS] Mag Status : 0
[2000-01-01 00:00:11][INFO] [4][COMMS] Checking transmit queue for packets to send...
[2000-01-01 00:00:11][INFO] [4][COMMS]   Transmit queue size: 0
[2000-01-01 00:00:11][INFO] [4][COMMS] Checking for incoming messages from GS...
[2000-01-01 00:00:11][INFO] [3][OBDH] Data processes: ['payload_tm', 'payload_od', 'eps_warning', 'payload_requests', 'hal', 'cdh', 'eps', 'cmd_logs', 'adcs', 'img']
[2000-01-01 00:00:12][INFO] [0][COMMAND] Popped command from queue: Command('ADCS_CTRL_MODE', id=31, args: ['mode_id']) with status: 0
[2000-01-01 00:00:12][INFO] [0][COMMAND]   Arguments: {'mode_id': 4}
[2000-01-01 00:00:12][INFO] Processing command: ADCS_CTRL_MODE with arguments: [4]
[2000-01-01 00:00:12][INFO] and precondition: valid_adcs_mode
[2000-01-01 00:00:12][ERROR] Cmd: Precondition failed
[2000-01-01 00:00:12][INFO] Added ack obj to transmit queue: Ack('rid=2', args=[31])
[2000-01-01 00:00:12][INFO] Command execution not successful due to error: 2
[2000-01-01 00:00:12][INFO] [0][COMMAND] Time: 946684812
[2000-01-01 00:00:12][INFO] [0][COMMAND] Time since boot: 12
[2000-01-01 00:00:12][INFO] [0][COMMAND] GLOBAL STATE: NOMINAL.
[2000-01-01 00:00:12][INFO] [0][COMMAND] RAM USAGE: 83%
[2000-01-01 00:00:12][INFO] [9][PAYLOAD] Payload state: OFF
[2000-01-01 00:00:12][INFO] [6][ADCS] Logging ADCS data at time 946684812
[2000-01-01 00:00:12][INFO] [6][ADCS] ADCS Mode : 3
[2000-01-01 00:00:12][INFO] [6][ADCS] Controller Mode : 0
[2000-01-01 00:00:12][INFO] [6][ADCS] Gyro Ang Vel : [0.00612526, -0.00239684, -0.00106526]
[2000-01-01 00:00:12][INFO] [6][ADCS] Mag Field : [7.45625e-05, -1.99375e-05, 1.5625e-06]
[2000-01-01 00:00:12][INFO] [6][ADCS] Sun Vector : [0.0, 0.0, 0.0]
[2000-01-01 00:00:12][INFO] [6][ADCS] Sun Status : 51
[2000-01-01 00:00:12][INFO] [6][ADCS] Gyro Status : 0
[2000-01-01 00:00:12][INFO] [6][ADCS] Mag Status : 0
```


flight/tasks/adcs.py
```diff
@@ -55,148 +61,222 @@ class Task(TemplateTask):
...
+        Total duration: 5 x 0.08 + 0.5 = 0.9 s.
+        """
+        self._mag_buffer = []
+        for k in range(self._MAG_N_SAMPLES):
+            status, reading = sensors.read_magnetometer()
+            if status == StatusConst.OK:
+                self._mag_buffer.append(reading)
+            if k < self._MAG_N_SAMPLES - 1:
+                TPM.sleep(self._MAG_SAMPLE_DT)
+
```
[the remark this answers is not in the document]
prchall123: design choice

flight/tasks/adcs.py
```diff
@@ -55,148 +61,222 @@ class Task(TemplateTask):
...
-            # Control MCMs and obtain coil statuses
-            dipole_moment = sun_pointing_controller(self.sun_pos_body, self.gyro_data, self.mag_data)
-        else:
-            # If in ACS_OFF or any other mode, do not control MCMs
-            # Just zero out the dipole moment
-            dipole_moment = np.zeros((3,))
+            elapsed = TPM.monotonic_float() - loop_start
+            remaining = GYRO_INTERVAL - elapsed
+            if remaining > 0:
+                TPM.sleep(remaining)
```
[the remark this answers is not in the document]
prchall123: design choice

flight/apps/adcs/consts.py
```diff
@@ -149,10 +213,10 @@ class MCMConst:
...
             [0.5, 0.0, 0.0],
-            [-0.5, 0.0, 0.0],
+            [0.5, 0.0, 0.0],
             [0.0, 0.5, 0.0],
-            [0.0, -0.5, 0.0],
+            [0.0, 0.5, 0.0],
+            [0.0, 0.0, 0.5],
             [0.0, 0.0, 0.5],
-            [0.0, 0.0, -0.5],
         ]
```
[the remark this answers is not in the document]
prchall123: their polarity is positive on either face

flight/apps/adcs/consts.py
```diff
@@ -57,13 +62,15 @@ class Modes:
...
 
-    SUN_VECTOR_REF = np.array([0.0, 0.0, 1.0])
-
     # Detumbling
-    TUMBLING_TOL = 0.09  # Exit detumbling into stable if ω < 0.09 rad/s (5 deg/s)
+    TUMBLING_TOL = 0.54  # Exit detumbling into stable if ω < 0.54 rad/s (30 deg/s)
+
+    # Detumbling only controllers
+    DETUMBLED_TOL_LO = 0.070  # Turn off detumbling  if ω < 0.07 rad/s (4 deg/s)
+    DETUMBLED_TOL_HI = 0.087  # Re-enter detumbling if ω > 0.087 rad/s (5 deg/s)
```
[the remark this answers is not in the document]
prchall123: difference between control target, and the detumbling vs nominal mode in fsw to give adcs focus

flight/tasks/watchdog.py
```diff
@@ -13,6 +14,9 @@ def __init__(self, id):
         self.name = "WATCHDOG"
 
     async def main_task(self):
+        if not CONFIG.ENABLED:
+            return
```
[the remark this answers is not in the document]
prchall123: file generated from the yaml files, it will have that configuration when the fsw is run

flight/apps/command/commands.py
```diff
@@ -501,7 +520,9 @@ def DELETE_ALL_FILES():
 
     try:
         DH.delete_all_files()
-        supervisor.reload()  # reload after deleting all files to clear any references to deleted files in memory and reset the state of the satellite
```
hazcoper, line 639: avoid unnecessary changes

flight/apps/command/processor.py
```diff
@@ -63,7 +63,10 @@ def process_command(command):
         elif not isinstance(response_args, (list, tuple)):
             response_args = [response_args]
 
-        return (CommandProcessingStatus.COMMAND_EXECUTION_SUCCESS, list(response_args))
```
hazcoper, line 66: avoid unnecessary changes

flight/apps/comms/comms.py
```diff
@@ -37,7 +37,7 @@ class SATELLITE_RADIO:
     undef_error_count = 0
     packet_none_count = 0
     packet_auth_fail_count = 0
-    rx_digipeater_count = 0    # the number of packets that match the digipeater header
```
hazcoper, line 40: avoid unnecessary changes

flight/apps/payload/controller.py
```diff
@@ -137,14 +137,17 @@ def switch_state(cls, state):
         if cls.current_state == PayloadState.IDLE:
             # set the last_executed_time
             cls.current_command = None
-            # TODO: maybe should change this not interesting to know that next command is at time  0, not enough resolution for it
```
hazcoper, line 140: avoid unnecessary changes

flight/apps/payload/download_manager.py
hazcoper, line 1: avoid unnecessary changes

flight/hal/argus_v4.py
hazcoper, line 1: avoid unnecessary changes

flight/tasks/comms.py
hazcoper, line 1: avoid unnecessary changes

flight/tasks/digipeater.py
hazcoper, line 1: avoid unnecessary changes

flight/tasks/payload.py
hazcoper, line 1: avoid unnecessary changes

flight/tasks/watchdog.py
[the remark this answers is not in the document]
hazcoper: correct, bue please implement this like it is implemented in other parts (add the value to a local variable and add the default value to satellite config (default value should be true))

flight/core/state_machine.py
hazcoper, line 1: avoid unnecessary changes

flight/core/states.py
hazcoper, line 1: avoid unnecessary changes

hazcoper (dismissed):

flight/configuration/flight.yaml
```diff
@@ -12,8 +12,8 @@ command:
     value: 18000  # seconds, i.e., 5 hours
     _const: true
 
-  PAYLOAD_TESTING_MODE:
-    value: false
+  SKIP_DEPLOYMENT:
```
ChaseDunaway21, line 15: I really don't like having something with the possibility of this being set to true. So at the very least we should note this in the checklist before integration. Otherwise I think we can remove this in flight.yaml and keep it in ground.yaml

flight/configuration/flight.yaml
```diff
@@ -71,3 +71,12 @@ log:
...
   LOG_FILE_LEVEL:
     value: "WARNING"
+
+# ADCS Task
+adcs:
+  CONTROLLER_MODE: 0
+
+# Watchdog Task
+watchdog:
+  ENABLED:
```
ChaseDunaway21, line 81: Same with this. Should be noted in the checklist!

flight/tasks/adcs.py
```diff
@@ -66,148 +76,156 @@ class Task(TemplateTask):
...
-                    else:
-                        zero_all_coils()
-
-                    self.mag_counter += 1
-                    if self.mag_counter == 5:
-                        self.mag_counter = 0
-
-            # Log data
-            # NOTE: In detumbling, most of the log will be zeros since very few sensors are queried
             self.log()
```
ChaseDunaway21, line 109: should the log happen before update_mode() or _bdot_cycle?

sil/fsw_plotter.py
```diff
@@ -247,38 +409,43 @@ def collect_FSW_data(outfile, result_folder_path, save_sil_logs=False, erase_sil
...
-
-        if adcs_mode:
-            adcs_modes.append([timestamp, int(adcs_mode.group(1))])
+
+        if adcs_timestamp:
+            current_adcs_time = float(adcs_timestamp.group(1))
+        elif adcs_mode:
+            adcs_modes.append([current_adcs_time, int(adcs_mode.group(1))])
         elif global_mode:
             global_modes.append([timestamp, global_mode_code[global_mode.group(1)]])
```
ChaseDunaway21, line 435: I think the timestamp will break if it isn't an integer epoch and is like YEAR-MONTH-DAY xxxx-xx-xx instead

flight/apps/adcs/consts.py
```diff
@@ -57,13 +62,15 @@ class Modes:
...
     ACS_OFF = 3  # Satellite has pointed to the sun and ACS can be turned off
 
-    SUN_VECTOR_REF = np.array([0.0, 0.0, 1.0])
-
     # Detumbling
-    TUMBLING_TOL = 0.09  # Exit detumbling into stable if ω < 0.09 rad/s (5 deg/s)
+    TUMBLING_TOL = 0.54  # Exit detumbling into stable if ω < 0.54 rad/s (30 deg/s)
+
+    # Detumbling only controllers
+    DETUMBLED_TOL_LO = 0.070  # Turn off detumbling  if ω < 0.07 rad/s (4 deg/s)
```
ChaseDunaway21, line 70: I'm somewhat confused by this These detumbling tolerance are now out of sync with the TUMBLING state tolerance (which is now 30deg/s). So will the state machine change from TUMBLING too early and trigger this tolerance immediately afterwards? This may not be an issue but I'm a little confused

flight/apps/adcs/sensors.py
```diff
@@ -30,8 +46,17 @@ def read_magnetometer() -> tuple[int, np.ndarray]:
     if SATELLITE.IMU_AVAILABLE:
         mag = 1e-6 * np.array(SATELLITE.IMU.mag())  # Convert field from uT to T
 
+        # mag validity check
+        is_valid = True
+        if mag is None or len(mag) != 3:
```
ChaseDunaway21: This if mag is None will never trigger if SATELLITE.IMU.mag() returns None since it will be converted to an array(None, dtype=object) first

flight/apps/adcs/acs.py
```diff
@@ -76,57 +77,97 @@ def sun_pointing_controller(sun_vector: np.ndarray, omega: np.ndarray, mag_field
...
+    # both coils on axis → each gets 0.5*val  (contributes 0.5*val^2 to norm^2)
+    # single coil on axis → it gets 1.0*val   (contributes val^2)
+    norm_sq = 0.0
+    for axis in range(3):
+        val = _U_EFF[axis]
+        if _COIL_STATUS[2 * axis] and _COIL_STATUS[2 * axis + 1]:
+            norm_sq += 0.5 * val * val
+        elif _COIL_STATUS[2 * axis] or _COIL_STATUS[2 * axis + 1]:
+            norm_sq += val * val
+    scale = max(1.0, norm_sq ** 0.5 * 0.707)
```
ChaseDunaway21, line 149: Is the power limit norm <= sqrt(2)? Shouldn't the L2 norm be enforcing <= 1? You already multiply norm_sq by 1/2 in the two coil case, then you multiply by it again pulled out of the root in the norm_sq ** 0.5 * 1/sqrt(2). I think you can simplify this by getting rid of the if elif statement and keeping the 0.707 factor in the max(...) statement. Or you could keep the if elif and get rid of the 0.707 factor. That's all to say if this should have a scale <= 1.

flight/apps/adcs/consts.py
```diff
@@ -74,85 +81,138 @@ class Modes:
...
+    @classmethod
+    def update_mode(cls, new_mode):
+        if new_mode in [cls.BDOT, cls.BCROSS, cls.SUN_POINTING]:
+            cls.current_mode = new_mode
+            try:
+                with open(_CTRL_MODE_PATH, "wb") as f:
+                    f.write(struct.pack("B", new_mode))
+                os.sync()
+            except Exception:
+                pass
```
ChaseDunaway21, line 136: Does this just return True no matter if it fails?

flight/hal/drivers/bmx160.py
```diff
@@ -625,27 +686,84 @@ def accel_powermode(self, powermode):
...
+    def _bmm150_write(self, bmm150_reg, value):
+        # Indirect write to BMM150 via BMX160 secondary interface (datasheet §2.4.3.1.1).
+        # Writing to MAG_IF[2] (0x4E) triggers the transaction; mag_man_op goes high
+        # while it is in progress and must clear before the next write is issued.
+        self.write_u8(_BMX160_MAG_IF_3_ADDR, value)
+        self.write_u8(_BMX160_MAG_IF_2_ADDR, bmm150_reg)
+        start_time = time.monotonic_ns()
+        while self.mag_man_op:
+            if time.monotonic_ns() - start_time > 1e6:  # 1 millisecond timeout
+                # TODO: Should throw an error
```
ChaseDunaway21: Just noting this again. currently it just passes without logging anything at the very least

flight/apps/adcs/sensors.py
```diff
@@ -30,8 +46,17 @@ def read_magnetometer() -> tuple[int, np.ndarray]:
     if SATELLITE.IMU_AVAILABLE:
         mag = 1e-6 * np.array(SATELLITE.IMU.mag())  # Convert field from uT to T
```
ChaseDunaway21, line 47: This is a second part to the BMX160 comment on breaking if the read fails, you can add a try catch here when calling SATELLITE.IMU.mag, but that may be unnecessary overhead. I'm not sure how much you've been able to test/seen an issue with this, it may completely be a non-issue.

flight/configuration/flight.yaml
prchall123: I understand the concern, though it applies to most of the variables in these config yaml files that serve to differentiate ground testing and flight build model, the checklist would have to go through them anyway. We could remove these options and have a ground testing checklist of where to implement these skips

flight/tasks/adcs.py
prchall123: I moved it to the start given the new actuation scheme. Given the uncertainty of the task scheduling, control is done entirely within each task call, so we can assume the coils will be off at the start. We log everything at the start and at about the same time and then proceed with the control and mode manager update. It is relevant for us that the measurements that get logged are correctly timestamped and unbiased from the coils for any ground sysid

sil/fsw_plotter.py
prchall123: the recent pr from saksham changed the time logging format to integer epoch to make it more compact

flight/apps/adcs/consts.py
prchall123: Sorry, should have documented this better. We have two thresholds for detumbling - one at 30 deg/s that tells the rest of the systems that we are tumbling too fast and need to give adcs priority/disable anything extraneous until detumbling has reached a safer level, and then another at 4/5 deg/s where we stop detumbling all together and turn the coils off because a) its low enough for us to do our operations b) gyro bias is in the order of 1 deg/s 1 sigma, so that 4 is a wide enough margin to cover it. 

It could be a design option to remove the 30 deg/s and only use the 4 deg/s limit instead, though that would complicate the spin-stabilized sun pointing a bit - it aims to rotate at 20 deg/s with the panels facing the sun instead of detumbling down to 4 deg/s.

flight/apps/adcs/acs.py
prchall123: The coils have a 25 Ohm resistance and a max voltage of around 6V, and u is a throttle from 0 to 1 being the max voltage. The max power of a coil is ~6^2/25 equal approx 1,44 W. 
The idea was to set the power limit at half the max power of a coil (around 0.7W) no matter which coils fail, only scaling down the control vector but not changing direction. Power is proportional to the current/voltage/throttle squared, so we get more efficient control per axis when divided between two coils than by a single one. That scale acts as a saturation function essentially, upper bounding the throttle with the sqrt of the max power. If we change from sqrt(P/2) to sqrt(P), we cap the power at 1,44 W.

flight/apps/adcs/sensors.py
prchall123: From what Ive seen, if the bmx160 fails, the mag and gyro functions will output all zeros, no part of the data collection process throws errors for this call to catch. adcs considers all zeros to mean the sensor isnt working and nothing should be done with those measurements

flight/apps/adcs/consts.py
prchall123: it doesnt return anything, the variable used to tell whether it has loaded or not is an attribute of the class. If it hasnt loaded an existing value successfully or created a new config file, it will stay loaded=false and try again next time

flight/apps/adcs/sensors.py
prchall123: In practice I haven't seen this happen because either one: the IMU boot sequence fails, an error is thrown and the device is None, which will be captured by the IMU Available check before this, or the IMU does boot up, but something weird happens with the sensor and the readings it outputs. In that case, the most likely scenario will be that the readings are exactly zero, which is something that we are guarded against after. 
I have added a check for it being an array with a None inside just in case, though we should keep in mind the minor trade-off for each check between bloating memory usage and runtime vs added safety

flight/hal/drivers/bmx160.py
prchall123: I've added a throw statement. This will only run during the boot sequence, if it throws an error IMU wont be available and ADCS wont run. This wont be called when we read the magnetometer.

flight/apps/adcs/consts.py
ChaseDunaway21: Ah, so at 30 deg/s, the ADCS has complete priority over the scheduler, then below that it operates as before until it reaches 4-5 deg/s? 

The sun-pointing makes sense, though it wouldn't take a pretty long time to spin up around the sun-pointing axis if it's tumbling at 30deg/s in possibly an arbitrary axis?

flight/apps/adcs/acs.py
ChaseDunaway21: Ah ok, that's makes sense, it just was a coincidence that the max power was approximately equal to 1/sqrt(2).

flight/apps/adcs/sensors.py
ChaseDunaway21: I guess you could check if it's a None-type before converting it to an array to keep the compute the same, but otherwise that also works

flight/apps/adcs/consts.py
prchall123: adcs always has priority, detumbling mode in fsw is a reference for other subsystems. The MC sims show how long it takes to sun point, in the order of a few hours

hazcoper (approved):

ChaseDunaway21 (approved):

PR #368, opened 2026-02-22
Gps bug fixes

### LINKED ISSUE #349: GPS Not Functioning
Under tests with both the SDR and the antenna board, GPS does not correctly return a valid nav message and never returns a valid position fix or timing fix.

ChaseDunaway21:
For both the S1216F8-GL (Flight Module) and PX1120S receivers, the GPS is now correctly functional. 

- Added automatic determination of receiver in use
- Removed previous unnecessary code
- Added helper functions for configuring the receivers (if previously unconfigured)
- Validated that the modules work in SDR testing
- Cleaned up the GPS Update function
- Fixed parsing float->int conversion

Note: Some values (such as lat/lon, SV's, etc.) are not available on the S1216F8-GL whereas they are on the PX1120S. Currently, the S1216F8-GL can get a fix and show proper ECEF, time, _dop, but will show 0 for number of satellites in view, lat/lon, and so forth.


flight/core/task_configuration.py
```diff
@@ -11,13 +11,13 @@
...
-    TASK.WATCHDOG: {"Task": watchdog, "Frequency": 1, "Priority": 1},
+    # TASK.WATCHDOG: {"Task": watchdog, "Frequency": 1, "Priority": 1},
     TASK.EPS: {"Task": eps, "Frequency": 5, "Priority": 2},
     TASK.OBDH: {"Task": obdh, "Frequency": 0.5, "Priority": 2},
     TASK.COMMS: {"Task": comms, "Frequency": 1, "Priority": 2, "ScheduleLater": True},
     TASK.ADCS: {"Task": adcs, "Frequency": 5, "Priority": 1},
-    TASK.GPS: {"Task": gps, "Frequency": 0.025, "Priority": 3, "ScheduleLater": True},
-    TASK.PAYLOAD: {"Task": payload, "Frequency": 1, "Priority": 3, "ScheduleLater": True},
+    TASK.GPS: {"Task": gps, "Frequency": 0.5, "Priority": 3, "ScheduleLater": True},  # GPS <= 1 Hz
+    TASK.PAYLOAD: {"Task": payload, "Frequency": 5, "Priority": 3, "ScheduleLater": True},
```
prchall123: why did this change?

flight/hal/drivers/gps.py
```diff
@@ -1,4 +1,18 @@
+"""
+Author: Chase
+Description: GPS Driver for the SkyTraq PX1120S and S1216F8-GL Modules
+
+This driver is designed to interface with the SkyTraq PX1120S and S1216F8-GL GPS modules over
+UART using their binary protocol. It handles board detection, message parsing, and data extraction for navigation information.
+
+Note, the PX1120S uses the AN0037 binary protocol, while the S1216F8-GL uses the AN0030 binary protocol.
```
prchall123: You should split the gps.py into PX1120S and S1216F8-GL classes and separate files. You can create an abstract gps superclass or not to avoid repeating code, though I would advise against it to keep it consistent with the rest of the drivers

flight/hal/argus_v4.py
```diff
@@ -99,9 +99,18 @@ class ArgusV4Interfaces:
...
     UART0_RX = board.RX0
-    UART0 = UART(UART0_TX, UART0_RX, baudrate=UART0_BAUD)
+    # UART0 = UART(UART0_TX, UART0_RX, baudrate=UART0_BAUD)
+
+    UART0 = UART(
+        UART0_TX,
+        UART0_RX,
+        baudrate=UART0_BAUD,
+        receiver_buffer_size=GPS_UART_RX_BUFFER_SIZE,
+    )  # TODO Tune this correctly
```
Mahii11: What is the status of this TODO?

flight/tasks/gps.py
```diff
@@ -115,12 +111,16 @@ async def main_task(self):
...
-                    self.log_info("GPS module did not send a valid nav data message")
+                    reason = SATELLITE.GPS.last_update_status
+                    if reason is None:
+                        reason = "GPS module did not send a valid nav data message"
+                    self.log_info(reason)
 
             else:
                 # GPS is not active in HAL
+                # print(f"GPS is not available: {SATELLITE.GPS_AVAILABLE}")
+                # print(f"GPS status: {SATELLITE.DEVICE_STATUS('GPS')}")
```
Mahii11: Delete these commented lines

ChaseDunaway21:
There is a temporary bug with the telemetry and data handler having a mismatch causing SIL to fail. I tested the fix by adding a placeholder cdh BOOT_TIME in datahandler and this seemed to resolve the issue.

flight/core/time_processor.py
```diff
@@ -62,6 +62,10 @@ class TimeProcessor:
     def calc_time_offset(cls):
         cls.time_offset = cls.time_reference - time.time()
 
+    @staticmethod
+    def _normalize_timestamp(unix_timestamp):
+        return int(unix_timestamp)
```
prchall123: minor pet peeve, but the code would be shorter and more readable by just using int directly over this function

flight/core/time_processor.py
```diff
@@ -104,7 +109,7 @@ def set_time(cls, unix_timestamp):
     def time(cls):
         if SATELLITE.RTC_AVAILABLE:
             # RTC exists, get RTC time
-            rtc_time = time.mktime(SATELLITE.RTC.datetime)
+            rtc_time = cls._normalize_timestamp(time.mktime(SATELLITE.RTC.datetime))
```
prchall123: I figured rtc time was already at second level resolution, why is this int needed now?

flight/tasks/gps.py
```diff
@@ -66,47 +65,44 @@ async def main_task(self):
...
-
                         DH.log_data("gps", self.log_data)
 
                         # If RTC is inactive, set the TPM time reference regardless of fix quality
                         if SATELLITE.RTC_AVAILABLE is False:
                             TPM.set_time_reference(SATELLITE.GPS.unix_time)
 
                         # Only update RTC time if the fix is better than _FIX_MODE_THR
-                        if SATELLITE.GPS.fix_mode > _FIX_MODE_THR:
+                        if SATELLITE.GPS.fix_mode >= _FIX_MODE_THR:
```
prchall123: maybe use an elif so you dont set the time twice if the rtc is not available

flight/tasks/gps.py
```diff
@@ -66,47 +65,44 @@ async def main_task(self):
         else:
```
prchall123, line 65: should we use the gps less in low power? is this something to be dealt with in a future pr?

flight/hal/drivers/px1120s.py
```diff
@@ -337,298 +470,329 @@ def is_leap_year(year):
                 total_days = 0
 
         # Convert days to seconds for Unix time calculation
-        days_since_unix_epoch = (year - 1970) * 365 + sum(is_leap_year(y) for y in range(1970, year))
+        days_since_unix_epoch = (year - 1970) * 365 + sum(self._is_leap_year(y) for y in range(1970, year))
```
prchall123, line 379: this approach might be off by the number of leap seconds between 1970 and 1980

flight/core/task_configuration.py
ChaseDunaway21: The payload frequency change must have been from something else, the GPS update frequency needs to increase in order to clear the buffer quicker and to reduce the odds of using stale GPS-nav data. I will revert the payload update frequency.

flight/core/time_processor.py
ChaseDunaway21: This function has become stale, but it was from a series of rewrites to try and get the unix timestamp to have higher precision from the GPS driver, though I later learned in testing that even as a double-precision float it didn't produce and additional precision beyond the decimal point.

flight/tasks/gps.py
ChaseDunaway21: The power change is a different PR

flight/hal/drivers/px1120s.py
ChaseDunaway21: It is accurate to the current leap seconds, it will be accurate until at least December per the IERS

prchall123: my bad, was thinking of unix as including leap seconds

flight/tasks/gps.py
ChaseDunaway21: Fixed

flight/core/time_processor.py
ChaseDunaway21: Fixed

flight/hal/drivers/gps.py
ChaseDunaway21: Fixed

flight/core/time_processor.py
ChaseDunaway21: Fixed

prchall123 (commented):
After this is tested on hardware and shown to be working with the recent changes, I'll approve

Mahii11 (dismissed):

prchall123 (approved):

NeilKhera (approved):
LGTM!

PR #378, opened 2026-03-24
chore(comms): removed some old code that is no longer used

hazcoper:


saksham2001 (dismissed):
Looks fine for the most part but I think you should also delete [get_tx_message_header](https://github.com/cmu-argus-2/FSW-mainboard/blob/e5a021599c3a587d7e3cc4702630fd63d6a6eaa5/flight/apps/command/commands.py#L349). Although that is not used anywhere.

saksham2001 (approved):

C15 (approved):

PR #387, opened 2026-03-31
Detumbling limit handling

prchall123:
## Pull Request for improving the handling of high tumble-rates

ADCS will have a limit to the angular rate it will be able to detumble from. This limit is mainly defined by the controller bandwidth, its update frequency and sensor sampling frequency. 

The ADCS control cycle runs entirely within the task, which occurs at a frequency of about 1 Hz. Within the cycle:

- B-dot first collects six measurements at a 12,5 Hz frequency in half a second, then uses them to compute b-dot with central finite diffing. The control is computed and applied for another half a second, taking up about one second to run the cycle. The two factors limiting how fast this algorithm can detumble from are the lag between the b-dot estimate and when the control is applied, and how frequently the controller is updated. For the former, the central difference approach with six measurements, used to make the bdot estimate less sensitive to the magnetometer noise, has the disadvantage of introducing more lag to the estimate, which will be about a quarter of a second before the control is applied. This could be reduced significantly by using a numerical approach that estimates the derivative at the end of the measurement cycle rather than in the middle, though that would increase the sensitivity to noise. Once control starts getting applied, the b-dot estimate is no longer updated to avoid a feedback loop between the magnetometers and the coils. The estimate could still be numerically propagated using the angular rate and last magnetometer/bdot estimate, but given the max detumbling rate is already 150  deg/s, which is fairly high, this feature was deemed unnecessary.
- B-cross collects the magnetic field and the angular rate and starts applying the control right after, without a dedicated measurement cycle. The angular rate is updated while the control is being applied for about a second. The magnetic field in the body frame could be rotated with the gyro and propagated forward to make the controller capable of detumbling from higher rates, but the current ceiling was considered high enough as well, and the feature was scrapped.

#### 1.1. New ADCS mode - Very Fast Tumbling

If the measured angular rate is past the point where it the controller will not work, it is best to not use it since a) as shown in sim, it will likely raise the angular momentum and worsen the problem and b) to save power. To this effect, a new mode called VF_TUMBLING, for very fast tumbling was created in which the controller is switched off. If the SC reaches this state, FSW should stop actuating.

#### 1.2. Ground control over Magnetometer Bias

The BMX160 magnetometer has a bias in the order of 40 muT, which renders its absolute value unusable. This means only b-dot will work uncalibrated. To be able to use b-cross or the sun pointing controller, the bias needs to be calibrated. Due to the limited memory available, no online calibration algorithm will be run. Instead, a collection of data from adcs is to be downlinked and used to estimate the bias on ground, after which a ground command can send the estimated bias to the satellite, enabling safe use of the sun pointing algorithm. The bias is stored in the sd card, same as the controller mode, such that it is persistent across reboots.

###### B-dot

B-dot detumbles in simulation up to about 190 deg/s. Helmholtz cage testing pointed to a max detumbling rate in between 150-160 deg/s, potentially pointing to some unaccounted lag between sim and hardware. The detumbling limit was set to 150 deg/s to be on the safe side. The results are outdated with the sim limit at 190 deg/s

image

###### B-cross

B-cross is no longer applied past 175 deg/s, where it was found to become unstable.

image

###### Sun Pointing

Sun pointing max tumble rate is similar to b-cross, since both run on a similar cycle at about 175 deg/s. 

image

image

#### Helmholtz Cage Testing

Only b-dot is tested in the Helmholtz cage. The phase between the control output and the cage sinusoid are compared, to see how much lag occurs between the spacecraft and the external b-dot. The phase crosses over 90 degree lag at between 150-160 deg/s, which is about 40 deg/s lower than expected from the MC sims.

image


flight/tasks/watchdog.py
```diff
@@ -4,22 +4,17 @@
...
     def __init__(self, id):
         super().__init__(id)
         self.name = "WATCHDOG"
-        self.enabled = CONFIG.ENABLED
 
     async def main_task(self):
-        if not self.enabled:
-            return
 
         if SATELLITE.WATCHDOG_AVAILABLE:
```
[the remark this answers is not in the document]
prchall123: intentional, disabling it this way will keep triggering it. Testing only config that won't be necessary anymore

flight/tasks/command.py
```diff
@@ -403,7 +405,9 @@ def state_machine_execution(self):
             self.log_info(f"PDMODE: {self.PAYLOAD_MODE}")
 
             """Transitions out of NOMINAL"""
-            if self.ADCS_MODE == Modes.TUMBLING and self.log_data[CDH_IDX.DETUMBLING_ERROR_FLAG] != 1:
+            if (self.ADCS_MODE == Modes.TUMBLING or self.ADCS_MODE == Modes.VF_TUMBLING) and self.log_data[
+                CDH_IDX.DETUMBLING_ERROR_FLAG
+            ] != 1:
                 # T2.1: Tumbling again AND detumbling error flag is not set, transition to DETUMBLING
```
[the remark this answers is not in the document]
prchall123: PR description has been updated, its ok to keep fsw in detumbling mode anyway

flight/tasks/command.py
```diff
@@ -377,7 +377,9 @@ def state_machine_execution(self):
                 self.log_data[CDH_IDX.DETUMBLING_ERROR_FLAG] = 1
 
             """Transitions out of DETUMBLING"""
-            if self.ADCS_MODE != Modes.TUMBLING or self.log_data[CDH_IDX.DETUMBLING_ERROR_FLAG] == 1:
+            if (self.ADCS_MODE != Modes.TUMBLING and self.ADCS_MODE != Modes.VF_TUMBLING) or self.log_data[
+                CDH_IDX.DETUMBLING_ERROR_FLAG
+            ] == 1:
                 # T1.1: Spin stabilized OR detumbling error flag is set, transition to NOMINAL
```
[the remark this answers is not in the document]
prchall123: description has been updated

flight/apps/adcs/consts.py
```diff
@@ -78,6 +81,50 @@ class Modes:
...
+    # def update_vf_tumbling_tols(cls, bdot, vf):
+    #     cls.VF_TUMBLING_TOL_BDOT = bdot
+    #     cls.VF_TUMBLING_TOL = vf
+    #     cls._save()
+    #
+    # @classmethod
+    # def update_detumbling_tols(cls, tumbling, lo, hi):
+    #     cls.TUMBLING_TOL = tumbling
+    #     cls.DETUMBLED_TOL_LO = lo
+    #     cls.DETUMBLED_TOL_HI = hi
```
ChaseDunaway21: This should be removed unless we plan to add a command to update these as well

flight/apps/adcs/consts.py
```diff
@@ -202,12 +240,73 @@ class ControllerConst:
...
+#     Constants used for magnetorquer control and allocation.
+#     """
+
 #     N_MCM = 6
 #     MCM_FACES = ("XP", "XM", "YP", "YM", "ZP", "ZM")
 #     MCM_INDICES = (0, 1, 2, 3, 4, 5)
+
 #     ALLOC_MAT = np.array(
 #         [
 #             [0.5, 0.0, 0.0],
```
ChaseDunaway21: Same with this

flight/apps/command/commands.py
```diff
@@ -611,6 +611,55 @@ def ADCS_CTRL_MODE(mode_id):
...
+#     Modes.update_vf_tumbling_tols(float(vf_bdot), float(vf))
+#     return [Modes.VF_TUMBLING_TOL_BDOT, Modes.VF_TUMBLING_TOL]
+#
+# @register_command()
+# def ADCS_UPDATE_DETUMB_TOLS(tb, dtb_lo, dtb_hi):
+#     """Updates the detumbling mode thresholds (exit-to-stable, turn-off, re-enter)."""
+#     logger.info(f"Executing ADCS_UPDATE_DETUMB_TOLS with tb_tol: {tb}, dtb_tol_lo: {dtb_lo}, dtb_tol_hi: {dtb_hi}")
+#     Modes.update_detumbling_tols(float(tb), float(dtb_lo), float(dtb_hi))
+#     return [Modes.TUMBLING_TOL, Modes.DETUMBLED_TOL_LO, Modes.DETUMBLED_TOL_HI]
+
```
ChaseDunaway21: Remove this unless we add the command

flight/tasks/adcs.py
```diff
@@ -95,6 +95,9 @@ async def main_task(self):
                 DH.register_data_process("adcs", _ADCS_DATA_FORMAT, True, data_limit=100000, write_interval=2)
 
             ControllerModes.load()
+            # ControllerConst.load()  # Removed: SD persistence for gains/inertia (RAM footprint)
+            # Modes.load()            # Removed: SD persistence for mode tolerances (RAM footprint)
```
ChaseDunaway21: Should be removed

ChaseDunaway21 (dismissed):
Everything looks good, you could remove the commented out blocks of other commands we could add later, but they are harmless to leave in.

flight/apps/adcs/sensors.py
```diff
@@ -10,6 +9,48 @@
...
+
+        try:
+            os.mkdir("/sd/config")
+        except Exception:
+            pass
+        with open(_MAG_BIAS_PATH, "wb") as f:
+            f.write(struct.pack(_MAG_BIAS_FMT, _MAG_BIAS[0], _MAG_BIAS[1], _MAG_BIAS[2]))
+        os.sync()
+    except Exception:
+        pass
```
hazcoper, line 52: this wont work great if we end changing low power mode to have the peripheral line off (sd card will not be powered)

hazcoper (commented):
Still need to test
There is the issue of using directly the sd card
We have to be careful with that if we end up turning of the peripheral line

flight/apps/adcs/sensors.py
prchall123: Right now the only thing that adcs does in low power is to log the measurements, if we go for that approach we would have to disable logging to the sd card in low power, measurements wouldn't be taken and this load could be skipped as well. I can have the load function not be called in low power just in case

hazcoper: yeah, not calling the load function and update_mag_bias would be great

prchall123: just updated so the load functions, logging and update arent called when the satellite is in low power

flight/tasks/adcs.py
prchall123: removed

flight/apps/command/commands.py
prchall123: removed

flight/apps/adcs/consts.py
prchall123: removed

flight/apps/adcs/consts.py
prchall123: removed

ChaseDunaway21 (approved):
Looks good, we should do a low power test and send the mag update command to check

hazcoper (approved):

PR #389, opened 2026-04-08
New python payload

hazcoper:
Added code to support the new python payload
Has all the necessary logic to perform capture images, perform inference and downlink the results

https://github.com/cmu-argus-2/FSW-Payload/pull/61
https://github.com/cmu-argus-2/GSW-backend/pull/48


flight/apps/payload/controller.py
```diff
@@ -31,878 +26,627 @@
...
+
+        # booting state needs to turn on the satellite
+        if cls.current_state == PayloadState.BOOTING:
+            logger.info("[PAYLOAD] -  Turning on jetson")
+            cls.log_data[PAYLOAD_IDX.LATEST_ERROR] = 200  # starting a new experiment resetting the error
+            cls.current_command = (
+                cls.get_first_command()
+            )  # choosing the command here to make sure that if boot fails we do not run the command again
+            cls.remove_first_command()
+            cls.turn_on_power()
```
Mahii11: Need to verify that this is working as expected

flight/hal/argus_v4.py
```diff
@@ -281,7 +280,9 @@ class ArgusV4Components:
     #########
 
     # JETSON
-    JETSON_UART = ArgusV4Interfaces.JETSON_UART
+    JETSON_UART = ArgusV4Interfaces.JETSON_UART    # TODO: i dont think we need this here
+    JETSON_SD_REQ = digitalio.DigitalInOut(board.JETSON_SD_REQ)   # this have been wired to 5v dcdc enable
```
Mahii11: Why is this changed?

flight/tasks/command.py
```diff
@@ -368,9 +366,9 @@ def state_machine_execution(self):
             # has access to the global state. External requests exists as a last resort to control the
             # payload from the CDH (and Payload task itself =/= Payload Controller)
 
-            # Note all ground commands related to the payload are executed in the command processor
-            if PC.state == PayloadState.READY:
-                pass
+            # # Note all ground commands related to the payload are executed in the command processor
+            # if PC.state == PayloadState.READY:
+            #     pass
```
Mahii11: Needs to be deleted before merge

flight/hal/argus_v4.py
```diff
@@ -106,7 +106,7 @@ class ArgusV4Interfaces:
     JETSON_BAUD = const(460800)
     JETSON_TX = board.TX1
     JETSON_RX = board.RX1
-    JETSON_UART = UART(JETSON_TX, JETSON_RX, baudrate=JETSON_BAUD, receiver_buffer_size=8192)
+    JETSON_UART = UART(JETSON_TX, JETSON_RX, baudrate=JETSON_BAUD, receiver_buffer_size=609 * 40, timeout=0.2)
```
Mahii11, line 109: Caution: need to keep an eye on this when doing DITL

flight/hal/argus_v4.py
hazcoper: @NeilKhera wired that pin into the enable of the 5v dcdc

flight/apps/payload/controller.py
Mahii11: Progress: working on daemon for running the script on Jetson.

Mahii11 (dismissed):

flight/apps/payload/controller.py
```diff
@@ -31,878 +26,627 @@
...
+        This should turn on power to the jetson
+        """
+        logger.debug("[PAYLOAD] Turning on Jetson power...")
+
+        if not SATELLITE.PAYLOADPOWER_AVAILABLE:
+            logger.warning("[PAYLOAD] Payload power pins is not available.")
+            return False
+
+        try:
+            SATELLITE.JETSON_ENABLE.value = False
```
prchall123: why is this set to false in the turn on function

flight/apps/payload/controller.py
```diff
@@ -31,878 +26,627 @@
...
+        Will send the command to turn off the jetson
+        """
+
+        # create the command
+        cmd_off = Command("TURN_OFF_PAYLOAD")
+
+        # send the command
+        PU.send(pack(cmd_off))
+
+        logger.error("Please implement me")
```
prchall123: we have functions to turn off the power and this method is called, does the above command not turn it off? why is this log here?

flight/apps/command/commands.py
prchall123, line 193: I know this was merged in a previous PR, but could we remove it here and only add it to main once the security risk is mitigated?

prchall123, line 349: again not from this PR, but would this list all be sent down in a single packet?

flight/apps/command/commands.py
```diff
@@ -85,6 +88,40 @@ def UPLINK_TIME_REFERENCE(time_reference):
...
+        SATELLITE.JETSON_SD_REQ.value = False  # turn of 5v dcdc to save more power
+
+    except Exception as e:
+        logger.error(f"[PAYLOAD] Failed to disable payload power: {e}")
+
+    return []
+
+
+@register_command()
+def TURN_ON_PAYLOAD():
```
prchall123, line 107: This is duplicated with the code in the controller.py, and its inconsistent. No point having two implementations of the same function

flight/apps/payload/controller.py
```diff
@@ -31,878 +26,627 @@
...
-    no_more_file_packet_to_receive = False
-    just_requested_file_packet = False
+    TELEM_TS = 0  # time at which last telemetry was requested
+    TELEM_PERIOD = 20  # request telemetry every 20s
 
-    # Batch transfer settings
-    USE_BATCH_TRANSFER = True  # Enable batch requests
-    BATCH_SIZE = 25  # Request 25 packets at a time
+    # Lets init uart connection.
+    # TODO should this only be made once the jetson has been turned on?
```
prchall123, line 97: Why hasn't this been addressed in this PR?

flight/apps/payload/controller.py
```diff
@@ -31,878 +26,627 @@
...
+    payload_tm_data_format = "QQQQ" + 14 * "B" + "H" + 2 * "B" + 3 * "H"
+    log_data = [0] * len(payload_tm_data_format)
 
-    # Communication statistics
-    crc_failure_count = 0
-    total_packets_received = 0
-    total_packets_retried = 0
-    packets_skipped_after_max_retries = 0  # Packets filled with zeros due to persistent CRC failures
+    # this is the dict were the transactions will be stored
+    # TODO - will probably only have one transaction at a time, might not be worth having a dict
```
prchall123, line 114: As far as I understood, the current payload only handles one transaction at a time. If this dict is unnecessary, please remove it along with this todo

flight/apps/payload/download_manager.py
```diff
@@ -0,0 +1,346 @@
...
+                raise Exception("No active transaction and queue is empty")
+
+        trans = self.current_transaction
+
+        # Listen for packets in this batch window
+        logger.info("Entering listen mode")
+        last_received_packet = TPM.time()  # this is the time at which the last received packet has been received
+        target_batch_count = self._calculate_batch_width()
+
+        while TPM.time() - last_received_packet < self.TIMEOUT_BETWEEN_PACKETS:
```
prchall123, line 166: This approach stalls the whole task scheduler, we need to make sure this can run safely in the latest mainboard rev without any issues, with a working watchdog

scripts/update_main.sh
prchall123, line 1: please remove this file from this pr

flight/apps/command/commands.py
```diff
@@ -85,6 +88,40 @@ def UPLINK_TIME_REFERENCE(time_reference):
...
+    """Sends a turn-on  command to the payload and turns off its power line."""
+    logger.info("Executing TURN_ON_PAYLOAD")
+
+    if not SATELLITE.PAYLOADPOWER_AVAILABLE:
+        logger.warning("[PAYLOAD] Payload power pins is not available.")
+        return ["payload power pins not available"]
+
+    try:
+        SATELLITE.JETSON_SD_REQ.value = True
+        time.sleep(0.1)
```
prchall123: once again, using sleep commands, even if only for 100 ms, adds risk. Please use the time processor sleep function and remove the time import. Add a comment acknowledging this risk

flight/apps/command/commands.py
```diff
@@ -85,6 +88,40 @@ def UPLINK_TIME_REFERENCE(time_reference):
...
     logger.info("Executing TURN_OFF_PAYLOAD")
+
+    if not SATELLITE.PAYLOADPOWER_AVAILABLE:
+        logger.warning("[PAYLOAD] Payload power pins is not available.")
+        return ["payload power pins not available"]
+
+    try:
+        logger.info("[PAYLOAD] Shutdown command sent successfully, waiting for payload to shutdown before cutting power")
+        SATELLITE.JETSON_ENABLE.value = False
+        time.sleep(0.1)
```
prchall123: same thing as comment below

flight/apps/command/commands.py
```diff
@@ -344,3 +378,71 @@ def UPDATE_SD_USAGE():
     except Exception as e:
         return [f"error: {e}"]
     return ["sd usage updated", usage]
+
+
+@register_command()
```
prchall123, line 381: we need to revise how experiments are setup. Its overkill to have many of these parameters configured from the mainboard/GS for every experiment. We can keep that separate from this PR

flight/apps/eps/eps.py
```diff
@@ -35,7 +32,7 @@ class EPS_POWER_THRESHOLD:
 def GET_EPS_POWER_FLAG(curr_flag, soc):
     """returns current EPS state based on SOC"""
     flag = EPS_POWER_FLAG.NONE
-    if EPS_POWER_FLAG.NONE <= curr_flag <= EPS_POWER_FLAG.EXPERIMENT:
+    if EPS_POWER_FLAG.NONE <= curr_flag <= EPS_POWER_FLAG.NOMINAL:
```
prchall123, line 35: We should still go into low power mode if we're in experiment mode and the power goes too low. We should handle that with a graceful shutdown

flight/apps/payload/controller.py
```diff
@@ -31,878 +26,627 @@
...
+        """
+        logger.debug("[PAYLOAD] Turning on Jetson power...")
+
+        if not SATELLITE.PAYLOADPOWER_AVAILABLE:
+            logger.warning("[PAYLOAD] Payload power pins is not available.")
+            return False
+
+        try:
+            SATELLITE.JETSON_ENABLE.value = False
+            time.sleep(0.1)  # TODO: probably do not need this delay
```
prchall123: TPM has a sleep function so you wouldnt need to import time. This method is redundant with the command. If this TODO is true, we should really dispose of unnecessary sleep calls

flight/apps/payload/controller.py
```diff
@@ -31,878 +26,627 @@
...
+            SATELLITE.JETSON_SD_REQ.value = False  # turn of 5v dcdc to save more power
+            logger.info("[PAYLOAD] Jetson power enabled successfully.")
+            return True
+        except Exception as e:
+            logger.error(f"[PAYLOAD] Failed to enable payload power: {e}")
+            return False
 
     @classmethod
     def turn_off_power(cls):
```
prchall123, line 666: same about this already being in commands

prchall123 (changes requested):

flight/apps/telemetry/splat
prchall123, line 1: test_comms and test_telemetry unit test files were previously removed with the reasoning that these could be added to the splat tests. If so, these should be added to splat, and the unit tests ci pipeline should still run these tests if relevant

flight/apps/command/commands.py
```diff
@@ -85,6 +88,40 @@ def UPLINK_TIME_REFERENCE(time_reference):
...
+    logger.info("Executing TURN_ON_PAYLOAD")
+
+    if not SATELLITE.PAYLOADPOWER_AVAILABLE:
+        logger.warning("[PAYLOAD] Payload power pins is not available.")
+        return ["payload power pins not available"]
+
+    try:
+        SATELLITE.JETSON_SD_REQ.value = True
+        time.sleep(0.1)
+        SATELLITE.JETSON_ENABLE.value = True  # TODO:Write a jetson available function in cubesat.py?
```
prchall123: was one not added in this PR?

flight/core/state_machine.py
```diff
@@ -66,7 +66,7 @@ def start(self, start_state=STATES.STARTUP):
         from core.task_configuration import TASK_CONFIG
 
         self.__task_config = TASK_CONFIG
-        self.__states = [STATES.STARTUP, STATES.DETUMBLING, STATES.NOMINAL, STATES.EXPERIMENT, STATES.LOW_POWER]
+        self.__states = [STATES.STARTUP, STATES.DETUMBLING, STATES.NOMINAL, STATES.LOW_POWER]
```
prchall123: Experiment mode shouldn't be removed - we want to keep a clear distinction between nominal and experiment

flight/core/states.py
```diff
@@ -13,8 +13,6 @@
     This state is active until stability is achieved or a timeout occurs.
     NOMINAL: The primary operational state where all systems, excluding payload, are fully functional,
     provided sufficient power levels are maintained.
```
prchall123, line 15: I don't agree with removing the experiment mode

flight/tasks/adcs.py
```diff
@@ -112,9 +112,9 @@ async def main_task(self):
                     self.MODE = Modes.STABLE
 
             # ------------------------------------------------------------------------------------------------------------------------------------
-            # LOW POWER or EXPERIMENT
+            # LOW POWER
```
prchall123, line 115: Again, I really don't agree with this

flight/tasks/payload.py
```diff
@@ -1,45 +1,35 @@
...
                 tag_name="payload_tm",
-                data_format=PC.tm_process_data_format,
+                data_format=PC.payload_tm_data_format,
                 persistent=True,
                 data_limit=100000,
                 circular_buffer_size=200,
             )
 
+        # TODO - not sure what this is for
         # OD process (should be a separate file process)
```
prchall123: this is something to be removed

prchall123 (changes requested):
I don't agree with removing the experiment mode.

flight/apps/payload/download_manager.py
Mahii11: I have done image downlink with this branch on the new flatsat builds, and I haven't experienced any issue with watchdog. Will keep an eye on it though.

flight/apps/command/commands.py
Mahii11: controller.py is internal to the payload, but the commands.py is from groundstation.

flight/apps/command/commands.py
hazcoper: yes, that is why you have the option for skip elements, to guarantee that you can see everything

flight/apps/command/commands.py
hazcoper: It is not a duplicate. 
But I think this command should be deleted before flight, mahlet added for debugging for now
There is no point in having a command to turn on the jetson, if you dont have other commands to do stuff with the jetson.

flight/apps/payload/controller.py
hazcoper: This is because as of right now at IST we do not have the carrier board to fully test if this should only be done with the jetson on or not.
But it is really irrelevant, potentially a very minor optimization

flight/apps/payload/controller.py
hazcoper: the effort to remove the dict is not worth with the gain of removing it.
the payload has already been fairly tested like this, I dont think we should change it now

flight/apps/payload/download_manager.py
hazcoper: Yeah, it has to be this way, otherwise the downloads would be extremely messy and slow.
It only happens for a few seconds when it is downloading the data.

flight/apps/command/commands.py
hazcoper: We do not have the tools to validate this. This was how @Mahii11 did it, so i kept it this way. will change to use TPM

flight/apps/command/commands.py
hazcoper: You can create a issue for this, but I think we have more pressing things to worry about then small optimisations

flight/apps/payload/controller.py
hazcoper: again, dont have the necessary things to validate that this can be taken out. It works like this, I dont think we should spend the time exploring if this could be removed

flight/apps/command/commands.py
hazcoper: Even if we dont have the extra firewall that was talked about. the command should stay.
if there are security concerns with this command, the solution should not be to erase it. The solution should be to not use it
It can stay there as a last resort if we end up needing it

flight/apps/eps/eps.py
hazcoper: I think that when entering in low power mode if there is an ongoing experiment it should be cancelled and the jetson should shutdown immediately  (not graceful).
currently this is how it is implemented

flight/apps/command/commands.py
prchall123: My point wasn't about command vs internal function, but the command using the internal function over having the same code in two places. If the command is only for debugging its fine

flight/apps/payload/controller.py
prchall123: can be addressed later

flight/apps/command/commands.py
Mahii11: @hazcoper yeah we can get rid of it,

prchall123 (approved):
we will need to test the fsw state machine changes, to be done in a separate issue and pr.

Mahii11 (approved):

PR #402, opened 2026-04-23
Telemetry update

hazcoper:
Commands added:
	Clear experiment list
		Clears all scheduled payload experiment commands currently stored in the payload command list. Use this to reset the payload schedule and stop any future queued experiment executions.

	Get experiment list
		Returns the current payload command list, including all scheduled experiment commands. Use this to verify what payload actions are queued and in what order.

	Get file size
		Given a file path, returns the size of that file. Useful for downlink planning, storage diagnostics, and validating file generation before transmission.

	Digipeater activate
		Enables digipeater operation so eligible packets can be relayed according to the configured relay rules.

	Digipeater deactivate
		Disables digipeater operation, preventing relay of incoming packets until re-enabled.

	RF resume
		Re-enables RF transmit capability after it has been stopped, allowing outbound acknowledgments and telemetry transmission again.

	RF stop
		Disables RF transmit capability to stop outbound transmissions. This is typically used for operational safety, test scenarios, or radio-control procedures.

	Ping
		Echo/health-check command used to verify command reception and end-to-end communications with the spacecraft.

	Pet reboot
		Resets the periodic reboot timer so the automatic reboot is postponed. This keeps the system running longer when uninterrupted operations are desired.

	Reboot with ACK
		Requests a deferred reboot that allows command acknowledgment handling before reboot execution. Designed so ground can receive confirmation before the restart occurs.

	Main power reboot
		Performs a reboot through the main spacecraft power/reboot path. Use when a full system restart is needed through the standard platform reboot mechanism.

	Graceful reboot
		Performs a controlled reboot sequence that attempts orderly shutdown actions first, reducing risk of data corruption and leaving subsystems in a cleaner state before restart.

	Get command list
		return a list with the splat commands and their arguments

Added implementations:
	rf stop
	digipeater
		fifo was modified to have a special behaviour

Removed:
	precondition
	satellite func
    ZP_SOLAR_CHARGE_VOLTAGE removed from dh constant
    ZP_SOLAR_CHARGE_CURRENT removed from dh constant
    ZM_SOLAR_CHARGE_VOLTAGE removed from dh constant
    ZM_SOLAR_CHARGE_CURRENT removed from dh constant
    removed old commands from splat
    removed scale from splat
    remove satellite field from command in splat
    argus id

Added:
	comms variables to telemetry
	added feature to have tasks that are not launched automatically
		they can be enabled/disabled
		used in digipeater
	Boot-time telemetry field addition.


Mahii11 (approved):
I'm skeptical on the changes to run.sh. I think it might be a good idea to leave it as it was to  make sure that people are able to compile for each board in the future if needed.

hazcoper:
The changes to run.sh are only removing the argus id variable that was being sent which is no longer used. There is no reason why it should not be compatible with other boards

flight/apps/comms/comms.py
```diff
@@ -118,6 +162,13 @@ def receive_message(cls):
 
         # hopefully we have a valid packet at this point
 
+        # Store raw bytes and feed digipeater queue before any validation
```
saksham2001, line 165: You added `cls.set_rx_mode()` in one of the commits "fixed issue where satellite was not able to receive two messages without transmitting" but then removed again. Just checking if removing it was intentional?

flight/apps/command/commands.py
```diff
@@ -52,10 +56,95 @@ def FORCE_REBOOT():
...
+        if not response:
+            logger.error("Failed to gracefully shutdown data handler, aborting reboot")
+            return ["graceful reboot failed: DH shutdown failed"]
+
+        SATELLITE.reboot()
+
+        return ["success"]  # this will never be returned
+    except Exception as e:
+        logger.error(f"Failed to gracefully reboot the satellite: {e}")
+        return ["graceful reboot failed: {e}"]
```
saksham2001, line 84: Not an f string will return the literal string: graceful reboot failed: {e}

saksham2001 (approved):

flight/apps/comms/comms.py
hazcoper: I can confirm that it is working now, It was a mistake I made previously

flight/apps/command/commands.py
hazcoper: ups, thanks. Will update in a another pr, not critical

PR #409, opened 2026-05-03
Od sat implementation

hazcoper:


flight/apps/command/commands.py
```diff
@@ -417,6 +563,7 @@ def EXPERIMENT(
     logger.info(f"[PAYLOAD] - Experiment command received to run at {ts}")
     result = PC.add_command(
```
ChaseDunaway21: I think this should be add_command_interface now


```suggestion
    result = PC.add_command_interface(
```

hazcoper: it should actually be add_command_inference
thanks for noticing

PR #426, opened 2026-05-16
Duty Cycle Radio and Turn Off GPS in LOW POWER Mode

### LINKED ISSUE #271: EPS: Turn GPS off and duty cycle radio in low power state
We need to disable certain components / take some measures in the low power state to ensure that we are in fact power positive.

- Kill power to GPS
- Toggle TX frequency of radio

ChaseDunaway21:
Tied to Issue:
EPS: Turn GPS off and duty cycle radio in low power state #271

We should be duty cycling the Radio and disabling the GPS in low power mode.


prchall123 (approved):

saksham2001 (approved):
Everything looks fine

PR #428, opened 2026-05-22
Od sat implementation v2

prchall123:


scripts/turn_jetson_on.py
```diff
@@ -23,3 +23,20 @@
...
+        The enable pin is used for a MOSFET to control the signal to the
+        MCU, the watchdog is powered regardless of the enable pin.
+        The input pin will therefore need to be toggled before the enable pin
+        is toggled to ensure that the watchdog is not triggered during the
+        transition.
+        """
+
+        if SATELLITE.WATCHDOG.input:
+            SATELLITE.WATCHDOG.input_low()
+        else:
```
saksham2001: this else is removed in watchdog.py but still kept here?

flight/tasks/watchdog.py
```diff
@@ -25,8 +25,7 @@ async def main_task(self):
 
             if SATELLITE.WATCHDOG.input:
                 SATELLITE.WATCHDOG.input_low()
-            else:
```
saksham2001, line 28: not sure about this, in this case it will be high->low->high, on low it won't change?

saksham2001 (dismissed):
I am not too clear with the watchdog changes, other things seem fine

flight/tasks/watchdog.py
prchall123: watchdog only pets in rising edge, if its high it needs to be set low to rise, if low it can just rise

scripts/turn_jetson_on.py
prchall123: good catch, haven't needed the script since the watchdog bug fix, I'll update it

saksham2001 (dismissed):

flight/apps/payload/controller.py
```diff
@@ -75,13 +75,15 @@ class PayloadController:
 
     # Time variables for diferente things
     BOOT_TS = 0  # time at which switched to booting state
-    BOOT_TIMEOUT = 60  # how long it will wait for jetson to respond ping
+    BOOT_TIMEOUT = 120  # how long it will wait for jetson to respond ping
```
ChaseDunaway21, line 78: We should experiment with this, make sure the jetson turns on almost guaranteed within 120...

prchall123: If it fails to boot the experiment throws that it failed at step 2 (booting), I've seen it happen less with 2 minutes, but have seen it happen still. However, not sure if there is another issue at play , because generally it boots in less than a minute

saksham2001 (dismissed):

saksham2001 (approved):

ChaseDunaway21 (approved):
I believe after our updates this should be good to merge.

PR #429, opened 2026-05-25
Ist gmsk new

hazcoper:
add support to switch to fsk
https://github.com/cmu-argus-2/FSW-mainboard/issues/423

it default to lora, there is a command that will allow to change to fsk with custom parameters
If there is a reboot, it will default back to lora

currently we are getting about 15kbit/s in fsk mode (compared to 4kbit/s in lora mode)

has been fully tested with modification to gs to use custom hardware. Finishing implementation in ground station to use the gs hardware

### DIFF
--- a/flight/apps/command/commands.py
+++ b/flight/apps/command/commands.py
@@ -239,8 +239,29 @@ def RF_RESUME():
 def SET_FSK(freq, power, br, ps, bandwidth, f_dev, p_len, p_detect, sync_len, ad_comp, pack_type, pl_len, crc_type, whitening):
     """Sets the radio modulation to FSK for the next transmission."""
     logger.warning("Executing SET_FSK: setting modulation to FSK for next transmission")
+    
+    from hal.configuration import SATELLITE
+    SATELLITE.RADIO.beginFSK(
+        freq=freq,
+        power=power,
+        bR=br,
+        pS=ps,
+        bW=bandwidth,
+        fDev=f_dev,
+        preLength=p_len,
+        preDetect=p_detect,
+        syncLength=sync_len,
+        addrComp=ad_comp,
+        packType=pack_type,
+        plLength=pl_len,
+        crcType=crc_type,
+        whitening=whitening,
+        currentLimit=140.0,
+        tcxoVoltage=1.7,
+        useRegulatorLDO=False,
+        blocking=True
+    )
 
-    SATELLITE_RADIO.set_modulation_fsk(freq, power, br, ps, bandwidth, f_dev, p_len, p_detect, sync_len, ad_comp, pack_type, pl_len, crc_type, whitening)
     return ["modulation_set_to_fsk"]
 
 
--- a/flight/apps/command/commands.py
+++ b/flight/apps/command/commands.py
@@ -156,6 +156,9 @@ def SUM(opA, opB):
     Test command
     used to experiment adding new command and testing the arguments
     """
+
+    SATELLITE_RADIO.set_modulation_gfsk()
+
     logger.info(f"Executing SUM with opA: {opA} and opB: {opB}")
     return [opA + opB]
 
--- a/flight/apps/command/commands.py
+++ b/flight/apps/command/commands.py
@@ -157,8 +157,6 @@ def SUM(opA, opB):
     used to experiment adding new command and testing the arguments
     """
 
-    SATELLITE_RADIO.set_modulation_gfsk()
-
     logger.info(f"Executing SUM with opA: {opA} and opB: {opB}")
     return [opA + opB]
 
@@ -235,6 +233,15 @@ def RF_RESUME():
     CommandSupervisor.cancel_pending_rf_stop()
     SATELLITE_RADIO.set_comms_mode(COMMS_MODE_ID.STANDARD)
     return ["rf_resume_executed"]
+
+
+@register_command()
+def SET_FSK(freq, power, br, ps, bandwidth, f_dev, p_len, p_detect, sync_len, ad_comp, pack_type, pl_len, crc_type, whitening):
+    """Sets the radio modulation to FSK for the next transmission."""
+    logger.warning("Executing SET_FSK: setting modulation to FSK for next transmission")
+
+    SATELLITE_RADIO.set_modulation_fsk(freq, power, br, ps, bandwidth, f_dev, p_len, p_detect, sync_len, ad_comp, pack_type, pl_len, crc_type, whitening)
+    return ["modulation_set_to_fsk"]
 
 
 @register_command()
--- a/flight/apps/command/commands.py
+++ b/flight/apps/command/commands.py
@@ -237,33 +237,42 @@ def RF_RESUME():
 
 @register_command()
 def SET_FSK(freq, power, br, ps, bandwidth, f_dev, p_len, p_detect, sync_len, ad_comp, pack_type, pl_len, crc_type, whitening):
-    """Sets the radio modulation to FSK for the next transmission."""
-    logger.warning("Executing SET_FSK: setting modulation to FSK for next transmission")
-    
-    from hal.configuration import SATELLITE
-    # SATELLITE.RADIO.beginFSK(
-    #     freq=freq,
-    #     power=power,
-    #     bR=br,
-    #     pS=ps,
-    #     bW=bandwidth,
-    #     fDev=f_dev,
-    #     preLength=p_len,
-    #     preDetect=p_detect,
-    #     syncLength=sync_len,
-    #     addrComp=ad_comp,
-    #     packType=pack_type,
-    #     plLength=pl_len,
-    #     crcType=crc_type,
-    #     whitening=whitening,
-    #     currentLimit=140.0,
-    #     tcxoVoltage=1.7,
-    #     useRegulatorLDO=False,
-    #     blocking=True
-    # )
-    SATELLITE.RADIO.beginFSK()
+    """
+    Sets the radio modulation to FSK for the next transmission
+    Calling the driver here directly to avoid passing a lot of arguments around
 
-    return ["modulation_set_to_fsk"]
+    for some reason the objects are wrapped in object wrapper
+    and object wrapper does not support __setattr__ so we are unable to change the values
+    to minimize changes, here we will use the obj attribute
+    doing all of this to avoid sending a lot of arguments in beginFSK
+    to avoid exhausting the stack
+    """
+    from hal.configuration import SATELLITE
+
+    logger.warning("Executing SET_FSK: setting modulation to FSK for next transmission")
+
+
+    SATELLITE.RADIO.obj._FREQ = freq
+    SATELLITE.RADIO.obj._POWER = power
+    
+    try:
+        SATELLITE.RADIO.beginFSK(
+            bR=br,
+            pS=ps,
+            bW=bandwidth,
+            fDev=f_dev,
+            preLength=p_len,
+            preDetect=p_detect,
+            syncLength=sync_len,
+            addrComp=ad_comp,
+            packType=pack_type,
+            plLength=pl_len,
+            crcType=crc_type,
+            whitening=whitening,
+        )
+    except Exception as e:
+        return [f"fsk failed {e}"]
+    return ["fsk set"]
 
 
 @register_command()
--- a/flight/apps/command/commands.py
+++ b/flight/apps/command/commands.py
@@ -252,6 +252,10 @@ def SET_FSK(freq, power, br, ps, bandwidth, f_dev, p_len, p_detect, crc_type, wh
 
     logger.warning("Executing SET_FSK: setting modulation to FSK for next transmission")
 
+    if not SATELLITE.RADIO_AVAILABLE:
+        logger.error("[COMMS ERROR] RADIO no longer active on SAT")
+        return ["radio not available"]
+
     SATELLITE.RADIO.obj._FREQ = freq
     SATELLITE.RADIO.obj._POWER = power
     SATELLITE.RADIO.obj._bitrate = br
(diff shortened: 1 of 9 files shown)


flight/apps/command/commands.py
```diff
@@ -234,6 +234,49 @@ def RF_RESUME():
...
+
+    will triplicate the tx burst size to make up for the faster transmission in fsk compared to lora
+
+    for some reason the objects are wrapped in object wrapper
+    and object wrapper does not support __setattr__ so we are unable to change the values
+    to minimize changes, here we will use the obj attribute
+    doing all of this to avoid sending a lot of arguments in beginFSK
+    to avoid exhausting the stack
+    """
+    from hal.configuration import SATELLITE
```
saksham2001, line 252: Like the other instances of this import, we should check SATELLITE.RADIO_AVAILABLE

flight/apps/command/commands.py
```diff
@@ -234,6 +234,49 @@ def RF_RESUME():
...
+    try:
+        SATELLITE.RADIO.beginFSK(
+            pS=ps,
+            bW=bandwidth,
+            fDev=f_dev,
+            preLength=p_len,
+            preDetect=p_detect,
+            crcType=crc_type,
+            whitening=whitening,
+        )
```
saksham2001, line 273: We should do some sort of basic validation for the arguments before switching to make sure the arguments are within valid ranges.

flight/apps/command/commands.py
```diff
@@ -234,6 +234,49 @@ def RF_RESUME():
...
+
+        # in fsk transmission is way faster compared to lora (>3 times)
+        # lets triplicate tx_burst_size
+        SATELLITE_RADIO.TX_BURST_SIZE *= 3
```
saksham2001: This is never reverted, if this command is send twice the TX_BURST_SIZE would be x9. We should either reset this value at reboot or make another command to get to LoRA. Earlier option would be easier. This should also be bounded.

flight/apps/command/commands.py
```diff
@@ -234,6 +234,49 @@ def RF_RESUME():
...
+
+    except Exception as e:
```
saksham2001, line 279: This won't work because if we try to change using beginFSK and get an error, now the radio is in an intermediate stage so it won't be able to return anything in LoRA or FSK. We should reboot here.

flight/apps/comms/comms.py
```diff
@@ -221,9 +231,14 @@ def transmit_message(cls, packet):
...
-            SATELLITE.RADIO.send(packet)
+            status = SATELLITE.RADIO.send(packet)
+
+            if status != 0:
+                logger.error(f"[COMMS ERROR] Failed to transmit packet, radio driver returned error code {status}")
+                cls.tx_failed_count += 1
+                return False
+
             cls.tx_packet_count += 1
-            logger.info(f"[COMMS] - Message has been transmitted: {format_bytes(packet[:20])}...")
```
saksham2001, line 226: no reason to remove this

flight/hal/drivers/sx126x.py
```diff
@@ -564,6 +566,47 @@ def begin(
...
+        self.setPacketParamsFSK(preLength, preDetect, 32, 0, 1, 255, crcType, whitening)
+
+        if useRegulatorLDO:
+            state = self.setRegulatorLDO()
+        else:
+            state = self.setRegulatorDCDC()
+
+        # 16 is the crc at the end, time in us
+        self._EXTRA_TIME = (self._preambleLength + self._syncWordLength + 16) / self._bitrate * 1000000
+
```
saksham2001, line 609: missing return statement at the end, consistent to all the beginX functions it should return state

flight/hal/drivers/sx126x.py
```diff
@@ -564,6 +566,47 @@ def begin(
...
+        self.setPacketParamsFSK(preLength, preDetect, 32, 0, 1, 255, crcType, whitening)
+
+        if useRegulatorLDO:
+            state = self.setRegulatorLDO()
+        else:
+            state = self.setRegulatorDCDC()
+
+        # 16 is the crc at the end, time in us
+        self._EXTRA_TIME = (self._preambleLength + self._syncWordLength + 16) / self._bitrate * 1000000
```
saksham2001, line 608: division by zero risk if something in beginFSK gives an error, make sure self._bitrate is not zero before doing this calculation?

flight/hal/drivers/sx126x.py
```diff
@@ -1542,17 +1742,17 @@ def setTxIq(self, txIq):
...
         self._preambleDetectorLength = preambleDetectorLength
-        if not self.blocking:
+        if not self._BLOCKING:
             ASSERT(super().startReceive())
 
-    def setBlockingCallback(self, blocking, callback=None):
-        self.blocking = blocking
-        if not self.blocking:
+    def setBlockingCallback(self, callback=None):
+        self._BLOCKING
```
saksham2001: no-op statement, should have been `self.blocking = blocking`?

flight/hal/drivers/sx126x.py
```diff
@@ -1605,27 +1801,21 @@ def _receive(self, len_=0, timeout_en=False, timeout_ms=0):
...
 
     def _transmit(self, data):
-        if isinstance(data, bytes) or isinstance(data, bytearray):
-            pass
-        else:
-            return 0, _ERR_INVALID_PACKET_TYPE
-
         state = super().transmit(data, len(data))
-        return len(data), state
+        return state
```
saksham2001, line 1819: to keep consistent with all the calls doing radio.send we should keep return len(data), state.

flight/hal/drivers/sx126x.py
```diff
@@ -1324,23 +1491,25 @@ def config(self, modem):
...
 
         data = [0, 0, 0, 0, 0, 0, 0]
-        data[0] = modem
+
+        self._modem = data[0] = modem
         state = self.SPIwriteCommand([_SX126X_CMD_SET_PACKET_TYPE], 1, data, 1)
         ASSERT(state)
 
-        data[0] = _SX126X_RX_TX_FALLBACK_MODE_STDBY_RC
+        data[0] = _SX126X_RX_TX_FALLBACK_MODE_FS
```
saksham2001, line 1501: The chip now keeps the PLL running between operations instead of dropping to standby, this would mean more current draw compared to standby?

flight/tasks/comms.py
```diff
@@ -54,23 +53,25 @@ async def transmit_message(self):
...
-            sent_in_burst += 1
-            if sent_in_burst >= self.TX_BURST_SIZE:
-                # Yield to scheduler after a burst so watchdog (and other tasks)
-                # get CPU time. Burst size bounded by HW watchdog timeout.
-                await sleep(0)
-                sent_in_burst = 0
+            SATELLITE_RADIO.set_tx_mode()  # set the radio to transmit mode if we have something to send, will be set back to rx mode after transmission
+            sent_in_burst = 0
+
+            while TransmitQueue.packet_available():
```
saksham2001, line 59: I think this should be now in try statement.

set_tx_mode() runs before the loop, set_rx_mode() after. If anything inside the loop raises (like division-by-zero on _bitrate=0, etc), the function exits with tx_en=True, rx_en=False. Every uplink missed until watchdog reset.

saksham2001 (changes requested):

flight/hal/drivers/sx126x.py
hazcoper: not relevant, immediately after config we go into receive mode
radio is never in standby

flight/apps/comms/comms.py
C15: This was removed to maximize radio bitrate. Logging here takes a long time and is costly during large downloads

flight/apps/command/commands.py
hazcoper: addressed in latest commit

flight/apps/command/commands.py
hazcoper: No need
the values are already checked at a driver level and raise exceptions if wrong
since we are in try except, and we added satellite reboot in except, no need for extra verifications

flight/apps/comms/comms.py
saksham2001: We are going to remove serial logging altogether, this should stay if we need to debug in SD logs

flight/apps/command/commands.py
hazcoper: addressed in latest commit

flight/apps/command/commands.py
hazcoper: addressed in latest commit

flight/apps/comms/comms.py
hazcoper: if we are able to downlink saved logs, there is no need for logging the messages we already received

flight/hal/drivers/sx126x.py
hazcoper: addressed in latest commit

flight/hal/drivers/sx126x.py
hazcoper: no need, it was already asserted as not zero in setModulationParamsFSK

flight/hal/drivers/sx126x.py
hazcoper: addressed in latest commit

flight/hal/drivers/sx126x.py
C15: The radio will always be used in blocking mode, and no calling functions are receiving this return. We suggest removing the len(data) as it is an additional space that will be used in the stack

flight/hal/drivers/sx126x.py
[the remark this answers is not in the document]
C15: addressed in the latest commit

flight/apps/command/commands.py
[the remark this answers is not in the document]
C15: addressed in the latest commit

flight/apps/command/commands.py
```diff
@@ -234,6 +234,49 @@ def RF_RESUME():
...
+        logger.error(f"FSK fail: {e}")
+        return [f"fsk failed {e}"]
+    return ["fsk set"]
```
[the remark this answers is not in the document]
C15: addressed in the latest commit. satellite reboots and radio reloads standard LoRa config

flight/tasks/comms.py
hazcoper: addressed in latest commit, comms app has internal track of radio state, if it tries to receive without being in rx mode, it will set rx mode

flight/hal/drivers/sx126x.py
```diff
@@ -564,6 +566,47 @@ def begin(
...
+
+        state = self.setCurrentLimit(currentLimit)
+        ASSERT(state)
+
+        state = self.setDio2AsRfSwitch(True)
+        ASSERT(state)
+
+        self.setWhiteningSeed()
+
+        self.setModulationParamsFSK(self._bitrate, pS, bW, fDev)
```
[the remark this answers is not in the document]
C15: Same as before, ASSERTS in the driver functions will cause the begin to fail and the set_fsk command function will reboot the satellite, returning it to LoRa

flight/apps/command/commands.py
```diff
@@ -234,6 +234,49 @@ def RF_RESUME():
...
+    doing all of this to avoid sending a lot of arguments in beginFSK
+    to avoid exhausting the stack
+    """
+    from hal.configuration import SATELLITE
+
+    logger.warning("Executing SET_FSK: setting modulation to FSK for next transmission")
+
+    SATELLITE.RADIO.obj._FREQ = freq
+    SATELLITE.RADIO.obj._POWER = power
+    SATELLITE.RADIO.obj._bitrate = br
```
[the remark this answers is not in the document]
C15: Same as before, ASSERTS in the driver functions will cause the begin to fail and the set_fsk command function will reboot the satellite, returning it to LoRa

prchall123 (approved):

flight/hal/drivers/sx126x.py
saksham2001: then why change this in the first place?

flight/apps/command/commands.py
saksham2001: it's still the same? I don't see this changed: `SATELLITE_RADIO.TX_BURST_SIZE *= 3`

hazcoper: SATELLITE_RADIO.TX_BURST_SIZE = 15 * br // 6000

it changed to this

flight/hal/drivers/sx126x.py
C15: The radio takes time to turn the OSC ON or OFF.

What happens in RX or TX with the STDBY_RC config:
- Packet RX or TX ends
- Radio module automatically changes mode to STDBY_RC, per config
- FSW sets the radio to RX or TX for the next packet
- FSW must wait for the OSC to initialize again before proceeding. The OSC was turned off, so this a "cold" start
- Packet RX or TX ends (the cycle repeats)

With the change we're proposing
- Packet RX or TX ends
- Radio module automatically changes mode to FS mode, per config
- FSW sets the radio to RX or TX for the next packet
- FSW must wait for the OSC to initialize again before proceeding. The OSC is on and tuned to the desired carrier frequency. Faster mode change
- Packet RX or TX ends (the cycle repeats)

The time the radio spends changing mode doesn't significantly impact power consumption, but it does make the link slower.

saksham2001 (approved):

flight/hal/drivers/sx126x.py
saksham2001: In this case the earlier comment was incorrect

PR #432, opened 2026-05-29
disabling the coils in Low Power Mode or when ADCS objective is achieved

### LINKED ISSUE #431: Set COIL_EN low when coils are not being used
Consuming ~120mW when coil drivers are idle

prchall123:
Magnetorquer coils consume ~0.1 W at rest. Disabling them lowers power consumption in low power mode by 15%. 

This PR has them disabled when in low power mode, or when adcs mode is acs_off, which is met when the sc has successfully detumbled or sun pointed, depending on the controller being used. The coils will still be kept enabled during the on/off cycling of the ADCS task to alternate between magnetometer measurement and magnetorquer actuation.

### DIFF
--- a/flight/hal/argus_v4.py
+++ b/flight/hal/argus_v4.py
@@ -812,6 +812,12 @@ class ArgusV4(CubeSat):
             return Errors.INVALID_DEVICE_NAME
         self.__turn_on_device(device_name)
 
+    def turn_off_device(self, device_name: str):
+        """turn_off_device: Turn off the device."""
+        if device_name not in self.__device_list:
+            return Errors.INVALID_DEVICE_NAME
+        self.__turn_off_device(device_name)
+
     def reboot(self):
         """Reboot the satellite by resetting the main power."""
         ArgusV4Power.MAIN_PWR_RESET.value = True
--- a/flight/tasks/adcs.py
+++ b/flight/tasks/adcs.py
@@ -4,6 +4,7 @@ import apps.adcs.sensors as sensors
 from apps.adcs.acs import (
     bcross_controller,
     bdot_controller,
+    enable_coils,
     mcm_coil_allocator,
     spin_stabilizing_controller,
     sun_pointing_controller,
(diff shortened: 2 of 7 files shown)


flight/hal/argus_v4.py
saksham2001, line 737: not related to this PR but should we not check device_cls.error and device_cls.device after the rebooting the coil if they are not working. This assumes that if we reboot they will start working.

flight/tasks/adcs.py
```diff
@@ -110,9 +112,9 @@ async def main_task(self):
 
                 self.log()
 
-            if SM.current_state == STATES.LOW_POWER or SM.current_state == STATES.EXPERIMENT:
+            if SM.current_state == STATES.LOW_POWER:
```
saksham2001, line 115: Just making sure - the PR description says disabling coils in Low power and Experimental mode. But here it's only low power mode.

saksham2001 (dismissed):
Minor comments

flight/tasks/adcs.py
prchall123: sc currently goes into experiment mode when it has a scheduled experiment. In case sun pointing works, we may want to keep it spin-stabilized sun pointed (at 20 deg/s) and then switch to detumbling (< 4 deg/s) when scheduling an experiment, and if it's in experiment mode that won't happen. We likely won't change experiment mode to be limited to when the Jetson is on at this point, so it's best kept this way.

flight/hal/argus_v4.py
prchall123: That check is already covered indirectly. After a power-cycle reboot, if the device still fails to initialize, SAMPLE_DEVICE_ERRORS in hal_monitor will detect DEVICE_NOT_INITIALISED on the next cycle (since device is None, dead and temp_disabled are both False) and feed back into handle_error, which increments error_count. Eventually error_count hits MAX_DEVICE_ERROR and the device is declared dead. Error counts also survive MCU resets via the SD card restore in hal_monitor's startup path.

saksham2001 (approved):

ChaseDunaway21 (approved):
I don't see any issues. If your testing has validated this, then I believe it's good to merge.

PR #434, opened 2026-05-31
Removed the serial logger

saksham2001:
In flight we do not need serial logging, we only need logging to the SD card. I have tried to remove the serial logger with very minimal changes. This should help save RAM and also make things (very slightly) faster.

Issue: https://github.com/cmu-argus-2/FSW-mainboard/issues/424


hazcoper (approved):

hazcoper (commented):
seems to not be logging anything to the sd card

PR #435, opened 2026-05-31
Digipeater fix

C15:
This PR changes how the digipeater looks for valid packets and how it creates the packets to digipeat.
As is, the code allows for a large variety of packets with invalid formats. With the implemented regex rules, the verification is stricter (although not perfect).

During testing, we saw no increase in RAM usage.

### DIFF
--- a/flight/apps/digipeater/aprs.py
+++ b/flight/apps/digipeater/aprs.py
@@ -18,29 +18,16 @@ def is_valid_lora_aprs_packet(data, re_obj):
     """Return True if data is a structurally valid LoRa APRS packet."""
 
     # check if there are enough bytes
-    if len(data) < _HEADER_LEN + 20:  # header + minimum APRS string length
+    if len(data) < _HEADER_LEN + 9:  # header + minimum APRS string length
         return 1
 
     # try an decode using ascii
     try:
         aprs_str = data[_HEADER_LEN:].decode("ascii")
     except (UnicodeDecodeError, ValueError):
-        return 3
+        return 2
 
-    # use re_obj to check if the satellite callsing is in the string
-    result = re_obj.search(aprs_str)
-    if result is None:
-        return 4
-
-    # ind max size: 10 (because of ssid)
-    # dst max size: 6
-    # extra chars: header + > = 4
-    # the maximum distance from the beggining of the string should be 20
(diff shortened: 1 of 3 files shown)


flight/tasks/digipeater.py
```diff
@@ -57,13 +80,14 @@ async def main_task(self):
...
-            if not result == 6:
-                self.log_warning(f"  Invalid packet format, dropping {result}")
+            result = is_valid_lora_aprs_packet(raw_packet, self._satellite_message_re)
+            if not result == 4:
+                self.log_warning(f"Invalid packet format, dropping {result}")
                 continue
 
             # Add asterik to callsign to indicate digipeating
-            final_packet = add_asterisk_packet(raw_packet, self._satellite_cs_re)
+            final_packet = add_asterisk_packet(raw_packet, self._satellite_replace_re,
```
saksham2001, line 89: raw_packet is bytes (raw radio data from comms.py:188), but _satellite_replace_re is compiled from a str pattern. re.sub will raise TypeError: cannot use a string pattern on a bytes like object on every valid packet, dropping the entire digipeater path.

flight/apps/digipeater/aprs.py
```diff
@@ -18,33 +18,21 @@ def is_valid_lora_aprs_packet(data, re_obj):
...
-    return 6
-
-
-def add_asterisk_packet(data, re_obj):
+def add_asterisk_packet(data, re_obj, callsign):
     """
     Will simply add an asterisk to the end of callsign in the path to indicate that the packed has been repeated
     """
-    return re_obj.sub(r"\g<0>*", data)
+    replacement = "\\1\\2\\3\\5\\6" + callsign + "*" + "\\8\\10"
```
saksham2001, line 37: I don't know if circuit python's re works with "\\10" for group 10.  just check to make sure it doesn't interpret it as group(1) + "0"

saksham2001 (changes requested):

flight/tasks/digipeater.py
C15: This didn't happen during our testing. As pictured below, only some of the packets weren't repeated. The valid ones were repeated as expected.

image

prchall123 (approved):

saksham2001 (approved):

